PR #14, opened 2018-01-29
Default documentation

bramtayl:
Might need testing I just did this visually

### DIFF
--- a/src/plugins/documenter.jl
+++ b/src/plugins/documenter.jl
@@ -67,6 +67,7 @@ function gen_plugin(
             end
            """)
     end
+end
 
 function show(io::IO, p::Documenter)
     spc = "  "


christopher-dG:
Looks good, thanks!
Please put the big string in a `text` variable like [here](https://github.com/invenia/PkgTemplates.jl/blob/75c7875899bdab4b05600f4766e2fbc70631e1ad/src/generate.jl#L209-L217), and then just `write(fp, text)`. The current style bugs me a bit for unknown reasons haha.
 [This test](https://github.com/invenia/PkgTemplates.jl/blob/75c7875899bdab4b05600f4766e2fbc70631e1ad/test/plugins/githubpages.jl#L35) is the only one that needs to be updated I think. Just testing something like `contains(index, "@autodocs\nModules = [$test_pkg]")` should be sufficient.

PR #19, opened 2018-09-24
Make PkgTemplates work on 1.0

christopher-dG:
This doesn't change much if any of the package functionality, it just makes it run and all the test pass. So it's basically creating packages for 0.6. Next up is to use `Pkg.generate` etc.  
One other change is with the default plugin file templates, there was a bug fixed in Mustache which lets us write much more clearly.


.travis.yml
```diff
@@ -3,17 +3,18 @@ os:
   - linux
   - osx
 julia:
-  - 0.6
```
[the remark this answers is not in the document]
iamed2: (and also require 0.7 instead of 1.0)

.travis.yml
```diff
@@ -3,17 +3,18 @@ os:
...
     - julia: nightly
   fast_finish: true
 notifications:
   email: false
-script:
-  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-  - julia -e 'Pkg.clone(pwd()); Pkg.test("PkgTemplates"; coverage=true)'
+before_script:
+  - git config --global user.name "Chris de Graaf"
+  - git config --global user.email "chrisadegraaf@gmail.com"
```
omus: Why is this needed?

defaults/appveyor.yml
```diff
@@ -36,8 +36,14 @@ build_script:
...
-  - C:\projects\julia\bin\julia -e "cd(Pkg.dir(\"{{PKGNAME}}\")); Pkg.add(\"Coverage\"); using Coverage; Codecov.submit(process_folder())"{{/CODECOV}}{{#COVERALLS}}
-  - C:\projects\julia\bin\julia -e "cd(Pkg.dir(\"{{PKGNAME}}\")); Pkg.add(\"Coverage\"); using Coverage; Coveralls.submit(process_folder())"{{/COVERALLS}}{{/AFTER}}
+after_test:
+{{#CODECOV}}
+  - C:\projects\julia\bin\julia -e "cd(Pkg.dir(\"{{PKGNAME}}\")); Pkg.add(\"Coverage\"); using Coverage; Codecov.submit(process_folder())"
+{{/CODECOV}}
+{{#COVERALLS}}
+  - C:\projects\julia\bin\julia -e "cd(Pkg.dir(\"{{PKGNAME}}\")); Pkg.add(\"Coverage\"); using Coverage; Coveralls.submit(process_folder())"
+{{/COVERALLS}}
+{{/AFTER}}
```
omus: The default appveyor.yml script should be now based upon the Appveyor.jl package.

.travis.yml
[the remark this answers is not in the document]
christopher-dG: 👍

defaults/appveyor.yml
christopher-dG: Yep, I'm going to change this in the next PR which will update the real functionality.

.travis.yml
christopher-dG: I should add a comment explaining. When you `LibGit2.commit` without these options, it raises an exception, and Travis OSX and AppVeyor both don't have this preconfigured (interestingly, it works on Linux). 
(A couple of commits are made as part of package generation)

omus: Ok, you may not want to expose your email address in the commit history. Up to you

christopher-dG: I mean, it's all over the commit history but I will remove it from the CI files 😅. Editing that commit.

PR #20, opened 2018-09-26
Generate packages for 1.0

christopher-dG:
Most of this is actually removing features that I don't consider useful enough to keep around.

Options to `Template` that have been removed:
- `years`: if you're generating a new project, you probably want the current year
- `requirements`: doesn't make sense since a goal of templates is reuse and requirements aren't a constant
- `gitconfig`: I don't think this was ever used

`force` is no longer an option for `generate`, it raises an error when the path already exists.

Plugin file templates have been updated, but I still need to test them.


iamed2:
I do think `requirements` can be useful but since it would have to change completely we can just implement it if it ever it becomes useful to us.

README.md
```diff
@@ -11,81 +11,143 @@ repeatable, and customizable way.**
...
 julia> t = Template(;
            user="myusername",
            license="ISC",
            authors=["Chris de Graaf", "Invenia Technical Computing Corporation"],
-           years="2016-2017",
            dir=joinpath(homedir(), "code"),
-           julia_version=v"0.5.2",
-           requirements=["PkgTemplates"],
-           gitconfig=Dict("diff.renames" => true),
+           julia_version=v"0.6",
```
omus: I'd use `v"0.7"` here

defaults/appveyor.yml
```diff
@@ -1,49 +1,33 @@
...
+  - julia_version: nightly
+platform:
+  - x86
+  - x64
 matrix:
   allow_failures:
-    - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x86/julia-latest-win32.exe"
-    - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x64/julia-latest-win64.exe"
-
+  - julia_version: nightly
```
omus: Should be indented once more

defaults/travis.yml
```diff
@@ -12,18 +12,18 @@ matrix:
...
 {{#AFTER}}
 after_success:
 {{#CODECOV}}
-  - julia -e 'cd(Pkg.dir("{{PKGNAME}}")); Pkg.add("Coverage"); using Coverage; Codecov.submit(process_folder())'
+  - julia -e 'using Pkg; Pkg.add("Coverage"); using Coverage; Codecov.submit(process_folder())'
 {{/CODECOV}}
 {{#COVERALLS}}
-  - julia -e 'cd(Pkg.dir("{{PKGNAME}}")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(process_folder())'
+  - julia -e 'using Pkg; Pkg.add("Coverage"); using Coverage; Coveralls.submit(process_folder())'
 {{/COVERALLS}}
```
omus: @ararslan I think you have a new an improved suggestion for this?

iamed2: Nope it hides test failures because it runs during the script not after.

ararslan: Heh, well, I assume you're referring to the new `codecov: true` and `coveralls: true`. @iamed2 and I noticed that it seems to mask test errors and report success. So perhaps best not to tout it just yet...

ararslan: Eric-Alex race condition

christopher-dG:
Currently testing at https://github.com/christopher-dG/TestPkg.jl
- Travis working
- AppVeyor TBD (queued)
- CodeCov working
- Coveralls integration is broken, but that's not our fault: https://github.com/JuliaCI/Coverage.jl/issues/175
- Documenter/GitHubPages working
- GitLabCI untested, I'll get around to that on the weekend hopefully

defaults/gitlab-ci.yml
```diff
@@ -1,45 +1,8 @@
...
-
 Julia {{VERSION}}:
   image: julia:{{VERSION}}
-  <<: *test_template
-
-Julia nightly:
-  image: staticfloat/julia:nightly-x64
-  allow_failure: true
-  <<: *test_template
+  script: julia --project='.' -e 'using Pkg; Pkg.build(); Pkg.test({{#GITLABCOVERAGE}}; coverage=true{{/GITLABCOVERAGE}})'
```
omus: Should this be `@.`? https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_PROJECT-1

christopher-dG: I suppose it could be, although there's basically a guarantee that there's a `Project.toml` in the current directory if the package was generated with `PkgTemplates`. I'll make that change though.

christopher-dG: Just waiting for https://gitlab.com/christopher-dg/TestPkg.jl/-/jobs/107653581 to pass

omus: Ok, besides this the PR looks great!

PR #29, opened 2018-10-22
Drop Git configuration test dependencies, add Test as test-only, use Pkg's devdir instead of our own

### LINKED ISSUE #25: Make Test a test-specific dependency
https://docs.julialang.org/en/v1/stdlib/Pkg/index.html#Test-specific-dependencies-1

### LINKED ISSUE #27: Use a local gitconfig for the test suite
JuliaCIBot does't have the git configurations we need (user.name, user.email) but that shouldn't be its responsibility. 

See the failing builds here: https://github.com/JuliaLang/METADATA.jl/pull/18839

### LINKED ISSUE #28: dev_dir should respect JULIA_PKG_DEVDIR
https://github.com/invenia/PkgTemplates.jl/blob/07e2277a17a86fff814d5a6b90f915f9e17db8a1/src/template.jl#L6
should use `JULIA_PKG_DEVDIR`, or call https://github.com/JuliaLang/Pkg.jl/blob/2039662ad21186767e58086f1cc349432bf2611e/src/Pkg.jl#L21 directly.

christopher-dG:
Closes #28 
Closes #27 (hopefully it solves the JuliaCIBot issues) 
Closes #25


src/generate.jl
```diff
@@ -30,7 +43,7 @@ function generate(pkg_name::AbstractString, t::Template)
         # Create the gh-pages branch if necessary.
         if haskey(t.plugins, GitHubPages)
             LibGit2.branch!(repo, "gh-pages")
-            LibGit2.commit(repo, "Empty initial commit")
+            LibGit2.commit(repo, "Initial commit")
```
omus: Is the initial commit no longer empty? The empty commit is useful as it makes diffs work better

src/generate.jl
```diff
@@ -298,9 +333,4 @@ function substitute(
     return substitute(template, merge(d, view))
 end
 
-"""
-    splitjl(pkg::AbstractString) -> AbstractString
-
-Remove ".jl" from the end of a package name if it is present.
-"""
```
omus: Why remove the docstring?

src/template.jl
```diff
@@ -47,22 +40,18 @@ create a template, you can use [`interactive_template`](@ref) instead.
         host::AbstractString="https://github.com",
         license::AbstractString="MIT",
         authors::Union{AbstractString, Vector{<:AbstractString}}="",
-        dir::AbstractString=dev_dir(),
+        dir::AbstractString=Pkg.devdir(),
```
omus: Just curious: is `~` supported here like it could be shown in the docstring?

src/generate.jl
christopher-dG: It's just being updated to match the initial commit message on master branch (they're both empty). I guess they both could be "Empty initial commit" but I find the wording of the former to be somewhat canonical.

src/generate.jl
christopher-dG: I got rid of the docstrings on non-exported functions mostly to get Documenter to stop yelling at me

src/template.jl
christopher-dG: Yes, `expanduser` is used

https://github.com/invenia/PkgTemplates.jl/blob/68ccca2ad6ffb121e8c4d40df6fcaa9d09297af5/src/template.jl#L75

src/generate.jl
christopher-dG: Also they're just oneliners and mostly self-explanatory

omus: Does this work for you? https://github.com/JuliaDocs/Documenter.jl/pull/172/files

omus: The nice part about keeping the docstrings is that they indicate that the function is intended to do. Without any docstrings or comments this information is lost.

christopher-dG: That option appears to only be for autodocs or am I wrong?

christopher-dG: One alternative could be to just put a `#` comment above the function.

PR #31, opened 2018-11-02
Adds additional kwargs to Documenter plugin

morris25:


src/plugins/documenter.jl
```diff
@@ -32,6 +32,19 @@ function gen_plugin(p::Documenter, t::Template, pkg_name::AbstractString)
...
 
+    kwargs_string = if :additional_kwargs in fieldnames(typeof(p))
+        set_kwargs = ["modules", "format", "pages", "repo", "sitename", "authors", "assets"]
+
+        # We want something that looks like the following:
+        #     key1="val1",
+        #     key2="val2",
+        #
+        kwargs = (x for x in p.additional_kwargs if first(x) ∉ set_kwargs)
+        join(string(tab, first(p), "=", repr(last(p)), ",\n") for p in kwargs)
```
omus: Split up pair earlier to make code easier to read:
```julia
(k for (k, v) in p.additional_kwargs if k ∉ set_kwargs)
```

src/plugins/documenter.jl
```diff
@@ -32,6 +32,19 @@ function gen_plugin(p::Documenter, t::Template, pkg_name::AbstractString)
         "[]"
     end
 
+    kwargs_string = if :additional_kwargs in fieldnames(typeof(p))
+        set_kwargs = ["modules", "format", "pages", "repo", "sitename", "authors", "assets"]
```
omus: Maybe rename to `standard_kwargs` or `ignored_kwargs`

src/plugins/documenter.jl
omus: There probably should also be a warning if a user accidentally included one of these keywords in the `additional_kwargs` field

test/tests.jl
```diff
@@ -13,6 +13,11 @@ end
 struct Bar <: CustomPlugin end
 # A dummy Plugin subtype.
 struct Baz <: Plugin end
+# A Documenter with extra kwargs
+struct Bat <: Documenter
```
omus: `Qux` is the next in line: https://en.wikipedia.org/wiki/Metasyntactic_variable

omus:
@christopher-dG any comments?

christopher-dG:
I'll review shortly

src/plugins/documenter.jl
```diff
@@ -32,6 +32,27 @@ function gen_plugin(p::Documenter, t::Template, pkg_name::AbstractString)
...
     end
 
+    kwargs_string = if :additional_kwargs in fieldnames(typeof(p))
+        standard_kwargs = ["modules", "format", "pages", "repo", "sitename", "authors", "assets"]
+
+        # We want something that looks like the following:
+        #     key1="val1",
+        #     key2="val2",
+        #
+        valid_keys = [k for k in keys(p.additional_kwargs) if k ∉ standard_kwargs]
```
christopher-dG: Any chance we could use `filter(k -> !in(k, standard_kwargs), keys(p.additional_kwargs))`?
I prefer to use map/filter over comprehensions and I'm not very fond of using unicode symbols.

src/plugins/documenter.jl
```diff
@@ -32,6 +32,27 @@ function gen_plugin(p::Documenter, t::Template, pkg_name::AbstractString)
...
+        if length(p.additional_kwargs) > length(valid_keys)
+            invalid_keys = (repr(k) for k in keys(p.additional_kwargs) if k in standard_kwargs)
+            @warn string(
+                "Ignoring predefined Documenter kwargs ",
+                join(invalid_keys, ", "),
+                " from additional kwargs."
```
christopher-dG: None of the log messages have punctuation, so maybe let's remove the period at the end here

test/tests.jl
```diff
@@ -397,4 +397,32 @@ end
...
+    pkg_dir = joinpath(t.dir, test_pkg)
+    p = Qux(
+        [],
+        Dict("strict"=>true, "checkdocs"=>:none, "format"=>:markdown, "stringarg"=>"string")
+    )
+    warn_str = "Ignoring predefined Documenter kwargs \"format\" from additional kwargs."
+    @test_logs (:warn, warn_str) gen_plugin(p, t, test_pkg)
+    make = readchomp(joinpath(pkg_dir, "docs", "make.jl"))
+    @test occursin(
+        strip("""
```
christopher-dG: Could we `lstrip` this string to make sure that the keyword list has a trailing newline?

src/plugins/documenter.jl
```diff
@@ -32,6 +32,27 @@ function gen_plugin(p::Documenter, t::Template, pkg_name::AbstractString)
...
+            )
+        end
+        join(string(tab, k, "=", repr(p.additional_kwargs[k]), ",\n") for k in valid_keys)
```
christopher-dG: I'd prefer `map(k -> string(tab, k, "=", repr(p.additional_kwargs[k]), ",\n"), valid_keys)` here too

src/plugins/documenter.jl
```diff
@@ -32,6 +32,27 @@ function gen_plugin(p::Documenter, t::Template, pkg_name::AbstractString)
         "[]"
     end
 
+    kwargs_string = if :additional_kwargs in fieldnames(typeof(p))
+        standard_kwargs = ["modules", "format", "pages", "repo", "sitename", "authors", "assets"]
```
christopher-dG: Perhaps make this a `const` at the top of the file? Also if we name it something like `standard_kws` if fits in 92 characters 🙃

christopher-dG:
One more thing we could do is check the `fieldtype` of `additional_kwargs` to make sure that it's not going to break.

test/tests.jl
```diff
@@ -397,4 +397,52 @@ end
...
+end
+
+@testset "Documenter add kwargs" begin
+    t = Template(; user=me)
+    pkg_dir = joinpath(t.dir, test_pkg)
+
+    function check_kwargs(kwargs, warn_str)
+        p = Qux(
+            [],
+            kwargs
```
christopher-dG: Trailing comma!

src/plugins/documenter.jl
```diff
@@ -32,6 +46,27 @@ function gen_plugin(p::Documenter, t::Template, pkg_name::AbstractString)
         "[]"
     end
 
+    kwargs_string = if in(:additional_kwargs, fieldnames(typeof(p))) &&
```
christopher-dG: `x in y` is a bit prettier

test/tests.jl
christopher-dG: actual upon further review this can definitely be on one line.

PR #33, opened 2018-11-05
Make Git stuff optional

### LINKED ISSUE #32: Can you create the package without it doing the "git" changes with branches, etc?
I would love to just be able to generate the templates and do the git stuff entirely myself.  Could that be an option?

christopher-dG:
Closes #32 

Still needs tests! And some behaviour can be changed a bit, e.g. we don't need to prompt for a host in interactive mode if git is set to false.


test/tests.jl
```diff
@@ -287,6 +287,19 @@ end
     rm(temp_dir; recursive=true)
 end
 
+@testset "Git-less template creation" begin
+    if isempty(LibGit2.getconfig("user.name", ""))
+        @test_nowarn Template(; user=me, git=false)
```
iamed2: `@test_nowarn` doesn't work with the new logging

test/tests.jl
```diff
@@ -27,14 +27,14 @@ const test_file = tempname()
...
 const template_text = """
-            PKGNAME: {{PKGNAME}}
-            VERSION: {{VERSION}}}
-            {{#DOCUMENTER}}Documenter{{/DOCUMENTER}}
-            {{#CODECOV}}Codecov{{/CODECOV}}
-            {{#COVERALLS}}Coveralls{{/COVERALLS}}
-            {{#AFTER}}After{{/AFTER}}
-            {{#OTHER}}Other{{/OTHER}}
-            """
+PKGNAME: {{PKGNAME}}
```
iamed2: This text block should be indented one level (along with the trailing triple-quote)

test/tests.jl
christopher-dG: Is the replacement for this `@test_logs do_thing_here()`?

christopher-dG: It seems to be working now, but I can't find any "official" information on how to do `@test_nowarn` in 1.0.

iamed2: Yeah there really isn't a good way to do it :/

iamed2:
What's with the invisible whitespace changes to licenses/LGPL-2.1+?

christopher-dG:
I did a grep for trailing whitespace and found that stuff, I don't know how it ever got there.

PR #35, opened 2018-11-05
Don't skip interactive mode tests on MacOS

christopher-dG:
Way back TerminalMenus.jl didn't work in Travis OSX tests, but apparently it's all good now. Also realized that using `@spawn` instead of `@async` is kind of useless.


omus:
Easier to read diff: https://github.com/invenia/PkgTemplates.jl/pull/35/files?w=1

christopher-dG:
> `?w=1`

Mind: Blown

Let me just fix that conflict really quick

christopher-dG:
Rebased

christopher-dG:
As a result of #31, defining `interactive` on any `Documenter` subtype isn't safe (as demonstrated by the broken tests) so I've moved it to `GitHubPages`.

PR #36, opened 2018-11-07
Fix UUID to match registry

iamed2:
The difference almost murdered my brain today by loading two versions of PkgTemplates in the same session.


christopher-dG:
Is the reason it's different here vs in the registry that it had a UUID automatically assigned to it when everything from METADATA originally got ported over to General?

iamed2:
TBH I have no idea why they're different

PR #41, opened 2018-11-09
Add new packages to the global environment

### LINKED ISSUE #40: Add new project to the (v1.0) environment
When used as a workflow for creating smaller projects, one thing that is confusing for introductory users is that the new package is not available by default in their global environment.

I think this is remedied by an`] NewPackage` after an `] activate` to ensure the global environment is selected. After that, I really like that it activates the new package manifest.

christopher-dG:
Closes #40 

I considered making this an option, but I'm trying to avoid adding too many because that gets hard to maintain. As mentioned in #40, this is the less surprising behaviour and experienced users will know how to remove it from the global env if they want to.


iamed2:
I think it would make more sense to `dev` the package in the currently activated environment

christopher-dG:
Ah I never really thought about that case. I think I just need to revert the env-activating logic in `gen_tests`.

christopher-dG:
There, now it uses the current env.

Issue #23, opened 2018-10-10
Rework Documenter plugin to use a separate project


christopher-dG:
https://discourse.julialang.org/t/psa-use-a-project-for-building-your-docs/14974

PR #42, opened 2018-11-12
Pin Documenter to v0.19

christopher-dG:
Just a temporary fix until #23 is addressed.


iamed2:
This only works if run on Julia 0.7+

christopher-dG:
Ah, I had not thought of that...

christopher-dG:
I think this should work now. Still messy but it's temporary.

PR #43, opened 2018-12-12
Update format in Documenter make script

mforets:
Just tried `PkgTemplates` for a new project, very cool! Building the docs gave me the following deprecation warning:

```julia
┌ Warning: `format = :html` is deprecated, use `format = Documenter.HTML()` instead.
```

With this change, probably [this test](https://github.com/invenia/PkgTemplates.jl/blob/62702da431f669f052aeb568718fcc373f00c0d2/test/tests.jl#L466) has to be updated as well. Using `Documenter v0.21.0` here.

### DIFF
--- a/src/plugins/documenter.jl
+++ b/src/plugins/documenter.jl
@@ -72,7 +72,7 @@ function gen_plugin(p::Documenter, t::Template, pkg_name::AbstractString)
 
         makedocs(;
             modules=[$pkg_name],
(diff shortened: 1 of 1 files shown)


christopher-dG:
This will actually break the existing Travis CI integration, because Documenter is pinned to an older version until I get around to properly implementing the new doc deployment workflow (#23).

PR #45, opened 2018-12-20
Drop AutoHashEquals dependency, other minor updates

christopher-dG:
I don't precisely remember why we needed AutoHashEquals (I think maybe Julia v0.x didn't implement `hash`, which `Dict` keys require, for new structs), but one test did still depend on it. Regardless, it's not needed anymore.

Other changes:
- Make struct fields concrete (mostly `AbstractString` -> `String`)
- Replace some big interpolated expressions with the varargs form of `print`

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1,8 +1,3 @@
-[[AutoHashEquals]]
-git-tree-sha1 = "45bb6705d93be619b81451bb2006b7ee5d4e4453"
-uuid = "15f4f7f2-30c1-5605-9d31-71845cf9641f"
-version = "0.2.0"
-
 [[Base64]]
 uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
 
--- a/Project.toml
+++ b/Project.toml
@@ -4,7 +4,6 @@ authors = ["Chris de Graaf <chrisadegraaf@gmail.com>"]
 version = "0.4.1"
 
 [deps]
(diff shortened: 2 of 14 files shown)


omus:
If your fields of your custom types don't use concrete types then you may need to change how hashing works. On Julia 1.0.3:
```julia
julia> using AutoHashEquals

julia> struct Foo1
           x::AbstractString
       end

julia> hash(Foo1("A")) == hash(Foo1(SubString("A")))
false

julia> @auto_hash_equals struct Foo2
           x::AbstractString
       end

julia> hash(Foo2("A")) == hash(Foo2(SubString("A")))
true

julia> struct Foo3
           x::String
       end

julia> hash(Foo3("A")) == hash(Foo3(SubString("A")))
true
```

christopher-dG:
Maybe I'm missing something, but I think that that's ultimately okay, since the keys of `Template.plugins` are the plugin types rather than values. 
Which would imply that AutoHashEquals was never actually required... I might have just gotten the idea that I needed when I wrote tests along the lines of `Set(values(t.plugins)) == Set([...])`.

omus:
I think you're safe to remove this now. I do think it's only safe to be removed because you also switched `AbstractString` -> `String`

christopher-dG:
I find it a lot easier to edit the parts of the string when a) it's properly syntax highlighted and b) there are less parens to deal with

PR #47, opened 2019-01-03
Add CirrusCI plugin for FreeBSD support

christopher-dG:
Uses Alex's [CirrusCI.jl](https://github.com/ararslan/CirrusCI.jl). The first and third commits are the only relevant ones, the second is just some things I found along the way.

### DIFF
--- a/defaults/cirrus.yml
+++ b/defaults/cirrus.yml
@@ -1,0 +1,16 @@
+freebsd_instance:
+  image: freebsd-12-0-release-amd64
+task:
+  name: FreeBSD
+  env:
+    JULIA_VERSION: {{VERSION}}
+  install_script:
+    - sh -c "$(fetch https://raw.githubusercontent.com/ararslan/CirrusCI.jl/master/bin/install.sh -o -)"
+  build_script:
+    - cirrusjl build
+  test_script:
+    - cirrusjl test
+{{#COVERAGE}}
+  coverage_script:
+    - cirrusjl coverage{{#CODECOV}} codecov{{/CODECOV}}{{#COVERALLS}} coveralls{{/COVERALLS}}
+{{/COVERAGE}}
--- a/src/plugins/cirrusci.jl
+++ b/src/plugins/cirrusci.jl
@@ -1,0 +1,45 @@
+"""
+    CirrusCI(; config_file::Union{AbstractString, Nothing}="") -> CirrusCI
+
+Add `CirrusCI` to a template's plugins to add a `.cirrus.yml` configuration file to
+generated repositories, and an appropriate badge to the README. The default configuration
+file supports only FreeBSD builds via [CirrusCI.jl](https://github.com/ararslan/CirrusCI.jl)
+
+# Keyword Arguments
+* `config_file::Union{AbstractString, Nothing}=""`: Path to a custom `.cirrus.yml`.
+  If `nothing` is supplied, no file will be generated.
+"""
+struct CirrusCI <: GenericPlugin
+    gitignore::Vector{String}
+    src::Union{String, Nothing}
+    dest::String
+    badges::Vector{Badge}
+    view::Dict{String, Any}
+
+    function CirrusCI(; config_file::Union{AbstractString, Nothing}="")
+        if config_file !== nothing
+            config_file = if isempty(config_file)
+                joinpath(DEFAULTS_DIR, "cirrus.yml")
+            elseif isfile(config_file)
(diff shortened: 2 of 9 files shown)


oxinabox:
mention it in the readme?

christopher-dG:
An overall summary of the available plugins up front and center would definitely be useful. This also reminds me I forgot to add it into the docs.

oxinabox:
Every other plugin I think is used in the example on
https://github.com/invenia/PkgTemplates.jl/pull/47/commits/4d5f767763d571420fac6ac11b6efdfcab0f7959#diff-04c6e90faac2675aa89e2176d2eec7d8R75

christopher-dG:
Oh right, I'll stick that in there too.

src/plugins/cirrusci.jl
```diff
@@ -0,0 +1,45 @@
...
+# Keyword Arguments
+* `config_file::Union{AbstractString, Nothing}=""`: Path to a custom `.cirrus.yml`.
+  If `nothing` is supplied, no file will be generated.
+"""
+struct CirrusCI <: GenericPlugin
+    gitignore::Vector{String}
+    src::Union{String, Nothing}
+    dest::String
+    badges::Vector{Badge}
+    view::Dict{String, Any}
```
omus: Just a thought: Can you leave out the `view` field from this plugin and instead use an accessor like: `view(::CirrusCI) = Dict{String, Any}()`?

christopher-dG: This has been on my mind for a while, just opened #48 to elaborate.

ararslan:
I'm flattered that you'd even consider adding this! Really my only concern is that this is _really_ new; I only got it semi-polished last night, and I don't think there are any packages currently using it. (Jackknife, SpecialFunctions, and BinaryProvider all have Cirrus CI set up, but they don't currently use my config.) It might be best to hold off for a bit until it's been put through its paces a bit more.

christopher-dG:
These one-config-file services are really easy to add so there was no reason not to :sweat_smile:.
Sounds good, we'll keep this open until then!

src/plugins/cirrusci.jl
```diff
@@ -0,0 +1,45 @@
+"""
+    CirrusCI(; config_file::Union{AbstractString, Nothing}="") -> CirrusCI
```
ararslan: A keyword argument that dictates the FreeBSD version to use might be nice. Unfortunately Cirrus doesn't document which are available, but I've successfully used the release images for 11.2 and 12.0. 12.0 seems like a good default, as the most recent release version of FreeBSD.

One could go even further and allow providing a vector of FreeBSD versions which would then turn into a `matrix` in the config file. That might be overkill though, since the changes between FreeBSD releases are unlikely to affect Julia code unless someone picks a really old FreeBSD version...

Anyway, just spitballing.

src/plugins/cirrusci.jl
ararslan: Also it doesn't seem to really be used anywhere. What's it for?

christopher-dG: Extra key-value pairs can go in there to be replaced in the template files.

oxinabox:
Can merge now?

defaults/cirrus.yml
```diff
@@ -0,0 +1,16 @@
...
+  name: FreeBSD
+  env:
+    JULIA_VERSION: {{VERSION}}
+  install_script:
+    - sh -c "$(fetch https://raw.githubusercontent.com/ararslan/CirrusCI.jl/master/bin/install.sh -o -)"
+  build_script:
+    - cirrusjl build
+  test_script:
+    - cirrusjl test
+{{#COVERAGE}}
```
ararslan: Coverage does not seem to work for Cirrus at all. I've opted to make this step always report success so as not to fail builds.

ararslan:
CirrusCI.jl has been around long enough and is now used in a slightly larger handful of packages that I think we could go forward with this.

christopher-dG:
@ararslan I can merge this now. Can you just confirm there are no changes required?

christopher-dG:
I've dropped some minor unrelated style changes from this PR which would be obliterated by #61 anyways.

christopher-dG:
Going ahead and merging this. @ararslan feel free to raise an issue if there is anything wrong.

ararslan:
Sorry, didn't get a chance to rereview. Seems good, thanks.

PR #50, opened 2019-01-10
Fix code loading in documentation generation

### LINKED ISSUE #49: PkgTemplates is not available during documentation generation
Docs for the stable and dev versions show:

image

christopher-dG:
Closes #49 

The examples run in a new environment since they create new packages, but PkgTemplates wasn't installed in it.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -23,6 +23,7 @@ run(`git config --global user.email "travis@c.i"`)
 run(`git config --global github.user "travis"`)
 using Pkg
 Pkg.activate(mktempdir())
+# This code gets run in docs/build/, so this path evaluates to the repo root.
 Pkg.add(PackageSpec(path=dirname(dirname(pwd()))))
 ```
 
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -23,6 +23,7 @@ run(`git config --global user.email "travis@c.i"`)
 run(`git config --global github.user "travis"`)
 using Pkg
 Pkg.activate(mktempdir())
+Pkg.add(PackageSpec(path=dirname(dirname(pwd()))))
 ```
 
 The simplest template requires no arguments.
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -18,9 +18,10 @@ pkg> add PkgTemplates
 ## Usage
 
 ```@setup usage
-run(`git config --global user.name "Travis"`)
-run(`git config --global user.email "travis@c.i"`)
-run(`git config --global github.user "travis"`)
+using LibGit2: getconfig
+isempty(getconfig("user.name", "")) && run(`git config --global user.name "Travis"`)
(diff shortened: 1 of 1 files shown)


iamed2:
Why not just `add` PkgTemplates instead of `dev`ing it?

christopher-dG:
Can I add a path? We want to make sure that the version installed is the version whose documentation is being built.

iamed2:
Yep!

```
  If a local path is used as an argument to add, the path needs to be a git repository. The project will then track
  that git repository just like if it is was tracking a remote repository online.
```

christopher-dG:
Ah, sweet. Is there a reason then why the recommended Documenter build stage uses `develop`?

iamed2:
Honestly I have no idea

christopher-dG:
🤷‍♂️

docs/src/index.md
```diff
@@ -23,6 +23,7 @@ run(`git config --global user.email "travis@c.i"`)
 run(`git config --global github.user "travis"`)
 using Pkg
 Pkg.activate(mktempdir())
+Pkg.add(PackageSpec(path=dirname(dirname(pwd()))))
```
omus: Should there be a comment indicating the expected directory structure?

christopher-dG: Yeah, probably

christopher-dG:
I've also added a small change to avoid globally configuring Git when building docs locally.

docs/src/index.md
```diff
@@ -18,11 +18,14 @@ pkg> add PkgTemplates
 ## Usage
 
 ```@setup usage
-run(`git config --global user.name "Travis"`)
-run(`git config --global user.email "travis@c.i"`)
-run(`git config --global github.user "travis"`)
+using LibGit2: getconfig
+isempty(getconfig("user.name", "")) && run(`git config --global user.name "Travis"`)
+isempty(getconfig("user.email", "")) && run(`git config --global user.email "travis@c.i"`)
+isempty(getconfig("github.user", "")) && run(`git config --global github.user "travis"`)
```
omus: Maybe the documentation should inform the user how to set this up properly?
- https://help.github.com/articles/setting-your-username-in-git/
- https://help.github.com/articles/setting-your-commit-email-address-in-git/

christopher-dG: Do you think it's necessary? The people building the docs locally are very likely to have already set this up.

omus: I'm not sure. Why do you have these instructions if we expect this is already setup?

christopher-dG: These aren't instructions, this code runs when documentation builds. The configuration is there because it's necessary for the examples to run (whose output goes into the documentation) and Travis builds don't configure Git for you. But the previous behaviour when building docs locally was to overwrite the user's Git configuration globally which was not ideal.

omus: lol, whoops I completely didn't notice that. I think I was viewing the rendered version. Nevermind then!

PR #51, opened 2019-01-22
Request: Sort lines in .gitignore file

tkf:
I know that this is kind of a bikeshedding, but let me request to sort lines in .gitignore.  I think it's better since it makes generated file more deterministic.

### DIFF
--- a/src/generate.jl
+++ b/src/generate.jl
@@ -229,12 +229,12 @@ Returns an array of generated file/directory names.
 """
 function gen_gitignore(pkg_dir::AbstractString, t::Template)
     pkg = basename(pkg_dir)
-    seen = mapfoldl(p -> p.gitignore, append!, values(t.plugins); init=[".DS_Store"])
-    if !t.manifest && "Manifest.toml" ∉ seen
-        push!(seen, "/Manifest.toml")  # Only ignore manifests at the repo root.
+    entries = mapfoldl(p -> p.gitignore, append!, values(t.plugins); init=[".DS_Store"])
+    if !t.manifest && !in("Manifest.toml", entries)
+        push!(entries, "/Manifest.toml")  # Only ignore manifests at the repo root.
     end
-    unique!(sort!(seen))
-    text = join(seen, "\n")
+    unique!(sort!(entries))
+    text = join(entries, "\n")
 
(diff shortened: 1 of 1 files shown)


src/generate.jl
```diff
@@ -229,14 +229,11 @@ Returns an array of generated file/directory names.
...
     pkg = basename(pkg_dir)
-    seen = [".DS_Store"]
-    t.manifest || push!(seen, "/Manifest.toml")  # Only ignore manifests at the repo root.
-    patterns = vcat(map(p -> p.gitignore, values(t.plugins))...)
-    for pattern in patterns
-        if !in(pattern, seen)
-            push!(seen, pattern)
-        end
+    seen = mapfoldl(p -> p.gitignore, append!, values(t.plugins); init=[".DS_Store"])
+    if !t.manifest && "Manifest.toml" ∉ seen
```
tkf: `"Manifest.toml" ∉ seen` part is actually unrelated to the main request.  But it gets rid of redundant `/Manifest.toml` entry in `.gitignore` when `Manifest.toml` is already listed.

src/generate.jl
```diff
@@ -229,14 +229,11 @@ Returns an array of generated file/directory names.
...
 function gen_gitignore(pkg_dir::AbstractString, t::Template)
     pkg = basename(pkg_dir)
-    seen = [".DS_Store"]
-    t.manifest || push!(seen, "/Manifest.toml")  # Only ignore manifests at the repo root.
-    patterns = vcat(map(p -> p.gitignore, values(t.plugins))...)
-    for pattern in patterns
-        if !in(pattern, seen)
-            push!(seen, pattern)
-        end
+    seen = mapfoldl(p -> p.gitignore, append!, values(t.plugins); init=[".DS_Store"])
```
christopher-dG: This can probably be renamed to `entries` now (more or less unrelated but imo it's more fitting).

christopher-dG:
I like this change.  Just one thing, could you replace `∉` with `!in`? I generally try to avoid Unicode in package code as a personal preference.

tkf:
Thanks for the code review. I included your suggestions in the patch.

christopher-dG:
Lovely, thank you!

PR #52, opened 2019-01-26
Use "dev" instead of "latest" to link GitHub pages

tkf:
It seems Documenter.jl changed the setting:
https://juliadocs.github.io/Documenter.jl/v0.21/lib/public/#Documenter.deploydocs


christopher-dG:
Thank you 😄

PR #54, opened 2019-01-27
Add GitLabPages plugin

tkf:
I was playing with GitLab pages and thought it would be nice if PkgTemplates can do what it does with `GitHubPages` plugin for GitHub.  So, this is my WIP try on implementing it.

I noticed that it's not straightforward to implement versioned documentation deploy like Documenter.jl does in Travis because it's coupled with how GitHub pages work (I think).  So, at the moment my PR only supports publishing `dev` (latest) documentation.  But I think the URL would be forward-compatible even after Documenter.jl supports GitLab CI.  So, before adding tests for it, I'd like know the feasibility for this PR to be merged even without versioned documentation support as in GitHub.

TODO:

- [x] Check that the project generated by this template actually works.  ~ATM, GitLab pages https://tkfm.gitlab.io/MyGitLabPlayground.jl/dev/ for my sample project https://gitlab.com/tkfm/MyGitLabPlayground.jl is 404. Apparently, this is a common problem in GitLab pages when it's first created: https://forum.gitlab.com/t/gitlab-pages-404-for-even-the-simplest-setup/5870~
- [x] Tests

### DIFF
--- a/src/plugins/gitlabpages.jl
+++ b/src/plugins/gitlabpages.jl
@@ -24,14 +24,10 @@ struct GitLabPages <: Documenter
 end
 
 function badges(::GitLabPages, user::AbstractString, pkg_name::AbstractString)
+    # We are only including a badge for `dev` documentation since versioned documentation
+    # is not supported in GitLab pages yet.  See:
+    # https://github.com/invenia/PkgTemplates.jl/pull/54
     return [
-        #=
-        format(Badge(
-            "Stable",
-            "https://img.shields.io/badge/docs-stable-blue.svg",
-            "https://$user.gitlab.io/$pkg_name.jl/stable"
-        )),
-        =#
         format(Badge(
             "Dev",
             "https://img.shields.io/badge/docs-dev-blue.svg",
--- a/src/plugins/gitlabpages.jl
+++ b/src/plugins/gitlabpages.jl
@@ -1,0 +1,51 @@
+"""
+    GitLabPages(; assets::Vector{<:AbstractString}=String[]) -> GitLabPages
+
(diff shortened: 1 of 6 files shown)


src/plugins/gitlabpages.jl
```diff
@@ -0,0 +1,51 @@
...
+            end
+        end
+        # Windows Git recognizes these paths as well.
+        new(["/docs/build/", "/docs/site/"], abspath.(assets))
+    end
+end
+
+function badges(::GitLabPages, user::AbstractString, pkg_name::AbstractString)
+    return [
+        #=
```
christopher-dG: How come this badge is commented?

tkf: I should remove `stable` badge completely or add some comment. As I wrote in the comment above https://github.com/invenia/PkgTemplates.jl/pull/54#issue-247934631, this PR only supports deploying `dev` document.

christopher-dG: Ah, thanks. I should have read more closely. 
I think removing the badge and adding a comment explaining why there's just one is best.

christopher-dG:
> I'd like know the feasibility for this PR to be merged even without versioned documentation support as in GitHub

I will gladly accept this PR without that support; master branch documentation is almost always sufficient.

src/plugins/gitlabpages.jl
tkf: Done in af79ad45db6513379886f195f782161cbc66eed7

tkf:
Thanks. I added some tests. I think it's good to go unless there are other problems.

christopher-dG:
The failing tests have nothing to do with these changes, they're caused by some change in Tables which is being addressed in a Mustache PR. Thank you so much!

PR #57, opened 2019-02-14
Clarify necessary Git options in docs

mortenpi:
First time user and I got hit with an error when I ran `t = Template()` because I didn't have `github.user` set and. The error only suggested passing an explicit argument to `Template`, so I think it would be good to clarify right away what it needs to work without arguments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ pkg> add PkgTemplates
 
 ## Usage
 
-The simplest template requires no arguments.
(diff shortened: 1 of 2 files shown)


christopher-dG:
LGTM. Could you please make the same change in `README.md`?

mortenpi:
Done! :slightly_smiling_face:

christopher-dG:
Perfect, thanks! :smile:

PR #61, opened 2019-03-01
Package overhaul

### LINKED ISSUE #48: Make Plugin interface less janky
Currently all the `GenericPlugin` subtypes have (and **must** have) some fields like `src`, `dest`, etc. This is not good for extension, so it would be nicer to have them implement methods like `source(::TravisCI)`, etc. This is especially nice on plugins which don't ever use certain fields like `gitignore` and `view`. `CustomPlugin`s also suffer to a lesser extent, as `gitignore` is a required attribute.
I think the current field-accessing behaviour still makes for an okay default though (i.e. `source(p::GenericPlugin) = p.src`).

### LINKED ISSUE #55: Simpler plugin type declarations
Most of the `GenericPlugin` subtypes look exactly the same except for a few word replacements, it would be nice to reduce the duplication even further.

Some macro could generate the basic code instead, and any extra functionality could be implemented afterwards. This will probably be easier after #48 is solved.

In this case, the distinction between `GenericPlugin` and `CustomPlugin` would probably no longer be necessary.

### LINKED ISSUE #72: Use Mustache for docs/ and test/?
It would be nice if it is possible to tweak templates for `docs/make.jl`, `docs/src/index.md`, and `test/runtests.jl`.

For example, I often [make `docs/make.jl` an "sandboxed executable"](https://github.com/tkf/Transducers.jl/blob/95520f0cb719827130eac7e912a7ec9d75e3e470/docs/make.jl#L1-L11) which makes the Travis setup as trivial as [`script: docs/make.jl`](https://github.com/tkf/Transducers.jl/blob/95520f0cb719827130eac7e912a7ec9d75e3e470/.travis.yml#L25).  Also, I want to wrap code in `test/runtests.jl` as a module to make the test reproducible in REPL.  I'd like to use those settings from the get-go.  However, it is not easy to do this at the moment.

### LINKED ISSUE #80: RFC: allowing to save template for later use?
It feels annoying that each time I create a template and I have to create it again after I exit REPL. It would be nice if we could have

- local template saving
- [github template repo](https://help.github.com/en/articles/creating-a-template-repository) support

A proposed workflow is to extend the `generate` function to accept a new keyword which would looks like

```
generate("MyPackage"; template="MyTemplate")
```

and for `interactive_template`, it would search a local folder to check if there is a previous saved template first.

we could have a `save` function that saves a generated template object to a folder inside `PkgTemplates` folder, or an external repo. I'm wondering if people would like this feature.

### LINKED ISSUE #87: Latest Travis CI changes : Windows + 32bit
Travis CI has added support for Windows and also for 32bit builds. It is useful if we update the template to include the latest improvements.
https://discourse.julialang.org/t/julia-is-now-available-on-travis-windows-builds/27187
https://discourse.julialang.org/t/32-bit-linux-julia-builds-are-now-available-on-travis-ci/27786

### LINKED ISSUE #92: GPG-sign the commits made by PkgTemplates.jl?
When PkgTemplates.jl creates a git repo, it makes two commits:
1. `Initial commit` - this commit is empty
2.  ` Files generated by PkgTemplates ` - pretty self-explanatory

Would it be possible to have those two commits automatically signed by my GPG key?

### LINKED ISSUE #95: Organization of test project should use `test/Project.toml`
Code at https://github.com/invenia/PkgTemplates.jl/blob/master/src/generate.jl#L141-L162

```julia
try
    Pkg.activate(pkg_dir)
    Pkg.add("Test")

    # Move the Test dependency into the [extras] section.
    toml = read(joinpath(pkg_dir, "Project.toml"), String)
    lines = split(toml, "\n")
    idx = findfirst(l -> startswith(l, "Test = "), lines)
    testdep = lines[idx]
    deleteat!(lines, idx)
    toml = join(lines, "\n") * """
    [extras]
    $testdep
    [targets]
    test = ["Test"]
    """
    gen_file(joinpath(pkg_dir, "Project.toml"), toml)
    Pkg.update()  # Regenerate Manifest.toml (this cleans up Project.toml too).
finally
    proj === nothing ? Pkg.activate() : Pkg.activate(proj)
end
```

does not follow the new organization of test project based on `test/Project.toml` from Julia v1.2

One does not need to split and shake so much the main project.toml starting from now.
`[extras]` will become obsolete in v2. 

See
https://julialang.github.io/Pkg.jl/v1/creating-packages/#Adding-tests-to-the-package-1

### LINKED ISSUE #96: Add GitHub Actions as a supported form of Continuous Integration (CI)
GitHub Actions will reach general availability on November 13. It would be great if, by that date, PkgTemplates was able to set up GitHub Actions for continuous integration (CI).

Unfortunately, I know very little about GitHub Actions. But I know that @davidanthoff has been working on Julia + GitHub Actions. Maybe he can provide some guidance.

### LINKED ISSUE #103: stop using `Pkg.generate`
`PkgTemplates.generate` calls `Pkg.generate`
https://github.com/invenia/PkgTemplates.jl/blob/26a494410a5617af35acd8097ed3bcf303d71543/src/generate.jl#L18-L19

It doesn't do much at all. It wouild be like 5 lines to replace

Two reasons to stop doing it:
1. Pkg developers often muse about just getting rid of `Pkg.generate` and telling people to use `PkgTemplates` 
2.  It prevents us from expanding on it easily. E.g. I wanted to make a PR just now to change it so that the main module included a module docstring.

### LINKED ISSUE #104: Override the default initial version number for the generated package
Currently, when you generate a new package with PkgTemplates, the generated package always has `version = "0.1.0"`.

It would be great if I could customize the initial version number.

Ideally, there would be a `initial_version::VersionNumber` keyword argument to `Template`.

christopher-dG:
Closes #72 
Closes #87 
Closes #95 
Closes #92
~~Closes # 90~~ 
Closes #80
Addresses #85 (does it supersede it? maybe)
Closes #103 
Closes #104 
Closes #96 

Rewrites are fun :man_shrugging: This got like, way out of hand.
It's finally ready for review!

In short, the new internal architecture makes plugins a lot more powerful and flexible than they used to be, so every part of package generation has been moved into various default plugins.

I've built the docs [here](https://cdg.dev/files/PkgTemplates/) so you don't need to build them yourself.

### TODO

- ~~Some more interactive code, it's pretty raw right now~~
- [x] More reference tests that use the full range of plugin options
- [x] Unit tests for the trickier bits of logic, ~~especially interactive stuff~~


christopher-dG:
This is getting way too big, I think I'll need to split into multiple PRs.

christopher-dG:
Note to self: use these for testing:

- https://github.com/Keno/TerminalRegressionTests.jl
- https://github.com/Evizero/ReferenceTests.jl

bramtayl:
@christopher-dG how can I help get this over the finish line? I have a new package I want to write

christopher-dG:
You could plan my math lessons for me so that I have time to code :sweat_smile:.
In seriousness, I'm pretty sure that this PR is more or less feature-complete, it just needs tests. What I really need is just a kick in the butt to find the time to finish it up, and hopefully this is it. Since it's conveniently Friday, I'll have a look at it this evening and see if I can at least re-familiarize myself with what exactly needs doing.

christopher-dG:
I looked over my work here this weekend and there's a whole lot that I want to (and have been) changing, but I can see myself getting this done in the next week or two, time permitting.

christopher-dG:
Title updated to reflect the somewhat significantly changed user APIs... :upside_down_face:

christopher-dG:
Wew, this is very close to ready (after a really really long time). The initial comment up top has been updated with the current state of things.

@iamed2 not sure how Invenia wants to go about reviewing this, I wouldn't really want to be tasked with looking at all this code myself... I'd be happy to walk someone through it over Slack or similar if that makes things easier. Just let me know what works best :slightly_smiling_face:

docs/src/developer.md
```diff
@@ -0,0 +1,320 @@
...
+    file::String = default_file("runtests.jl")
+end
+
+source(p::Tests) = p.file
+destination(::Tests) = joinpath("test", "runtests.jl")
+view(::Tests, ::Template, pkg::AbstractString) = Dict("PKG" => pkg)
+
+function hook(p::Tests, t::Template, pkg_dir::AbstractString)
+    # Do the normal BasicPlugin behaviour to create the test script.
+    invoke(hook, Tuple{BasicPlugin, Template, AbstractString}, p, t, pkg_dir)
```
nickrobinson251: why `invoke` instead of just a normal function call? e.g.

```julia
basic_hook(t::Template, pkg_dir::AbstractString) = ...
hook(p::BasicPlugin, t, pkg_dir) = basic_hook(t, pkg_dir)
function hook(p::Tests, t, pkg_dir)
    # Do the normal BasicPlugin behaviour to create the test script.
    basic_hook(t, pkg_dir)
    # Do some other work.
    add_test_dependency()
end
```

docs/src/migrating.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Plugins
+
+Aside from renamings, basically every plugin has had their constructors reworked.
+So if you are using anything non-default, you should consult the new docstring.
+
+| Old           | New                    |
+| :-----------: | :--------------------: |
+| `GitHubPages` | `Documenter{TravisCI}` |
+| `GitLabPages` | `Documenter{GitLabCI}` |
```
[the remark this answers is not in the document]
nickrobinson251: oh it's just the TravisCI is the only gitub-compatible CI we have a plugin for right now. Maybe comment that underneath?

src/template.jl
```diff
@@ -1,242 +1,138 @@
+default_plugins() = [ProjectFile(), SrcDir(), Git(), License(), Readme(), Tests()]
+default_user() = LibGit2.getconfig("github.user", "")
```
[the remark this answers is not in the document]
nickrobinson251: I think no, because of how plugins use the `user` field e.g. it is interpolated into the travis url

README.md
```diff
@@ -1,146 +1,39 @@
...
 
-**PkgTemplates is a Julia package for creating new Julia packages in an easy,
-repeatable, and customizable way.**
+**PkgTemplates creates new Julia packages in an easy, repeatable, and customizable way.**
 
-## Installation
+Assuming you have the relatively standard Git options `user.name`, `user.email` and `github.user` set up globally with `git config --global`, creating a `Template` is as simple as:
 
-```julia
-pkg> add PkgTemplates
```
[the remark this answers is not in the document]
christopher-dG: Yeah, we can. I get tired of writing the same instructions everywhere but I forget that this package is commonly used by newcomers.

.gitignore
```diff
@@ -1,6 +1,6 @@
+/docs/build/
+/docs/site/
 .DS_Store
```
Nosferican: This is Mac OS specific. For OSX users, they should globally git ignore `.DS_Store` files as best practices.

docs/src/developer.md
christopher-dG: I'd prefer to not tell people to call an internal function, this way it's explicit that they're replicating the existing behaviour (at least it is to me :upside_down_face:).

docs/src/migrating.md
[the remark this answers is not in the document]
christopher-dG: Yeah, if we ever have something like docs via GitHub Actions or similar, there would be `Documenter{GitHubActions}`. Whether that actually ends up making sense will kind of depend how Documenter evolves to handle deployment from various CI providers. 
IMO there's no more clarification required here, but the `Documenter` docstring is really vague so that needs some improving.

docs/src/migrating.md
```diff
@@ -0,0 +1,65 @@
...
+| :-----------------------------------------: | :---------------------------------: |
+| `generate(::Template, pkg::AbstractString)` | `(::Template)(pkg::AbstractString)` |
+
+## Interactive Templates
+
+Two less names to remember!
+
+| Old                    | New                                 |
+| :--------------------: | :---------------------------------: |
+| `interactive_template` | `Template(; interactive=true)`      |
```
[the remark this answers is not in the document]
christopher-dG: Yeah I'm still working on this bit.

docs/src/migrating.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Other Functions
+
+Another two less names to remember!
+Although it's unlikely that anyone used these.
+
+| Old                  | New                                                                                 |
+| :------------------: | :---------------------------------------------------------------------------------: |
+| `available_licenses` | [GitHub](https://github.com/invenia/PkgTemplates.jl/tree/master/templates/licenses) |
+| `show_license`       | [GitHub](https://github.com/invenia/PkgTemplates.jl/tree/master/templates/licenses) |
```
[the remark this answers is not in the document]
christopher-dG: Ah, the link should say something more like "view licenses on GitHub", because there is no longer any equivalent function in PkgTemplates. Does that help the confusion?

src/template.jl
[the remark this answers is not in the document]
christopher-dG: Yeah, `user.name` is a real-world name i.e. Nick Robinson whereas we need a username i.e. nickrobinson251 here.

docs/src/index.md
```diff
@@ -1,107 +1,29 @@
...
-        AppVeyor(),
-        GitHubPages(),
-        CirrusCI(),
-    ],
-)
-generate("MyPkg2", t)
-run(`git -C $(joinpath(t.dir, "MyPkg2")) ls-files`);
-```
+This documentation refers plenty to [`Template`](@ref)s, the package's main type.
+But it also refers to "template files" and "text templating", which are plaintext files with placeholders to be filled with data, and the technique of filling those placeholders with data, respectively.
```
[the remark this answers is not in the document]
christopher-dG: Agree. Good idea on the `note` box.

.gitignore
christopher-dG: Yeah I wasn't sure what to do for that, it is included in the gitignore by default on older PkgTemplates versions and now it's an option to the `Gitignore` plugin. Fair to remove it from both I suppose?

docs/src/developer.md
christopher-dG: In general I tend to think "the smaller the API surface, the better"

README.md
```diff
@@ -1,146 +1,39 @@
...
 [![Build Status](https://travis-ci.org/invenia/PkgTemplates.jl.svg?branch=master)](https://travis-ci.org/invenia/PkgTemplates.jl)
-[![Build Status](https://ci.appveyor.com/api/projects/status/r24xamruqlm88uti/branch/master?svg=true)](https://ci.appveyor.com/project/christopher-dG/pkgtemplates-jl/branch/master)
 [![Codecov](https://codecov.io/gh/invenia/PkgTemplates.jl/branch/master/graph/badge.svg)](https://codecov.io/gh/invenia/PkgTemplates.jl)
 
-**PkgTemplates is a Julia package for creating new Julia packages in an easy,
-repeatable, and customizable way.**
+**PkgTemplates creates new Julia packages in an easy, repeatable, and customizable way.**
 
-## Installation
+Assuming you have the relatively standard Git options `user.name`, `user.email` and `github.user` set up globally with `git config --global`, creating a `Template` is as simple as:
```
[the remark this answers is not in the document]
christopher-dG: Thanks for the feedback here, good ideas.
About the exception timing, I think this is enough to justify adding one more function `validate(::Template, ::Plugin)` that runs in the `Template` constructor. 
There's other similar validation stuff that runs in prehooks like the `BasicPlugin` prehook but it would be better to have it done at constructor time.

docs/src/migrating.md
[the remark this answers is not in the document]
nickrobinson251: yeah, that helps

docs/src/developer.md
nickrobinson251: hmm, i think of `invoke` as a scary thing to be avoided. And using it is one thing, but having it as the main recommended way of extending functionality doesn't feel great to me. That said, I see my objection here is largely a personal preference

.gitignore
nickrobinson251: why remove it? seem very low cost to have (a little bit useful to some people, basically zero cost to everyone)

christopher-dG: I am fine with leaving it in this package's gitignore, it makes no difference to me. And turns out what I mentioned about the Git plugin does not actually make any sense since I removed the specific `ignore_ds_store` option and just left the open-ended `ignore` list option.

christopher-dG:
Thanks for the reviews @nickrobinson251! I have addressed each point except I am just working on the `validate`concept right now. Let me know if the updates need more updating.

docs/src/developer.md
christopher-dG: Aw, I really like `invoke` :sweat_smile:. I am going to keep it as is for now, but the second that another person says something along the lines of "using `invoke` for this is bad/awkward" I will change it.

An alternative would be to do the `invoke`ing internally, i.e. check if there is a specific method implemented for a concrete type, and if so, call both the ::Plugin method (via `invoke`) and the ::Subtype method (via a normal call). But that does not allow you to actually disable the default, so I don't really like it.

.gitignore
Nosferican: I think so. Mac users should really have it in their global gitignore.

christopher-dG:
Squashed the first ~50 or so commits of which there are few remnants in the current codebase. I'm comfortable with not squashing this PR now, since i'm pretty sure that each commit should at least compile.

.gitignore
christopher-dG: I removed it, I like the idea of this package having the same gitignore that would be generated by default.

src/plugin.jl
```diff
@@ -1,153 +1,161 @@
...
-            else
-                throw(ArgumentError(
-                    "File \$(abspath(config_file)) does not exist"
-                ))
+struct Foo <: Plugin
+    file::String = "/the/default/value" <- "This is the interactive prompt"
+    n::Int <- "This one has no default, so it's a required keyword"
+    abc::String = "No prompt, so the default is always taken in interactive mode"
+    xyz::String  # Required keyword, with a generic interactive prompt.
+end
```
nickrobinson251: This example does not work (at 441eee3)

```julia
julia> using PkgTemplates

julia> PkgTemplates.@with_defaults struct Foo <: Plugin

           file::String = "/the/default/value" <- "This is the interactive prompt"

           n::Int <- "This one has no default, so it's a required keyword"

           abc::String = "No prompt, so the default is always taken in interactive mode"

           xyz::String  # Required keyword, with a generic interactive prompt.

       end
ERROR: LoadError: type String has no field args
Stacktrace:
 [1] getproperty(::Any, ::Symbol) at ./Base.jl:20
 [2] (::PkgTemplates.var"##21#23"{Expr,Array{Expr,1}})(::Expr) at /Users/npr/repos/PkgTemplates.jl/src/plugin.jl:37
 [3] foreach(::PkgTemplates.var"##21#23"{Expr,Array{Expr,1}}, ::Array{Any,1}) at ./abstractarray.jl:1920
 [4] @with_defaults(::LineNumberNode, ::Module, ::Expr) at /Users/npr/repos/PkgTemplates.jl/src/plugin.jl:25
in expression starting at REPL[2]:1
```

christopher-dG: Oops. This is kind of half-baked, I'm contemplating moving it to a new InteractiveStructs package.

christopher-dG:
I have made the decision to wait until after this PR to do the interactive stuff. For whatever reason I am having a hard time coming up with code that I like for it and I just want to get this across the finish line. 

At this point I'd say this is just pending review.

src/template.jl
```diff
@@ -1,242 +1,143 @@
...
+        foreach(plugins) do p
+            if needs_username(p)
+                T = nameof(typeof(p))
+                s = """$T: Git hosting service username is required, set one with keyword `user="<username>"`"""
+                throw(ArgumentError(s))
+            end
         end
     end
+
+    isempty(kwargs) || @warn "Unrecognized keywords were supplied" kwargs
```
aminya:
```suggestion
    isempty(kwargs) || @warn "Unrecognized keywords were supplied. If you are using the old syntax, please see [Migrating To PkgTemplates 0.7+](migrating.md)." kwargs
```

docs/src/migrating.md
```diff
@@ -0,0 +1,61 @@
...
+\* `develop=true` was the default setting, but it is no longer the default in PkgTemplates 0.7+.
+
+## Plugins
+
+Aside from renamings, basically every plugin has had their constructors reworked.
+So if you are using anything non-default, you should consult the new docstring.
+
+| Old           | New                    |
+| :-----------: | :--------------------: |
+| `GitHubPages` | `Documenter{TravisCI}` |
```
aminya:
```suggestion
| `GitHubPages` | `Documenter{TravisCI}()` |
```

docs/src/migrating.md
```diff
@@ -0,0 +1,61 @@
...
+| `GitLabPages` | `Documenter{GitLabCI}` |
```
aminya:
```suggestion
| `GitLabPages` | `Documenter{GitLabCI}()` |
```

docs/src/user.md
```diff
@@ -0,0 +1,239 @@
...
+
+Using PkgTemplates is straightforward.
+Just create a [`Template`](@ref), and call it on a package name to generate that package:
+
+```julia
+using PkgTemplates
+t = Template()
+t("MyPkg")
+```
+
```
aminya:
```suggestion

A simple example would be:
```julia
using PkgTemplates

t = Template(
    dir = pwd(),
    user = "YourID",
    authors = ["Your Name"],
    julia = v"1.0",
    plugins = [

        License(name = "MIT"),
        Documenter{TravisCI}(),
        Develop(),
        TravisCI(),
        AppVeyor(),
        Codecov(),
    ],
)


t("PackageName")
```
```

**Comment by aminya** [comment]
Something strange happens when I try to publish a repository that is generated by this package:
image

File: docs/src/migrating.md (hunk 12 above)
Comment (christopher-dG) [reply docs/src/migrating.md]: This is inaccurate, the table is a mapping from type to type, so putting a constructor call on one side doesn't mean the same thing.

File: docs/src/user.md (hunk 14 above)
Comment (christopher-dG) [reply docs/src/user.md]: I will add something similar but further down. I didn't want to show all the options and plugins before explaining them.

**Comment by christopher-dG** [comment]
Did you create a repository on GitHub before trying to push to it?

File: docs/src/migrating.md (hunk 12 above)
Comment (aminya) [reply docs/src/migrating.md]: I think the user doesn't necessarily know that these are types. If you provide an example however, that resolves the confusion.

**Comment by aminya** [comment]
> Did you create a repository on GitHub before trying to push to it?

In Github Desktop, there is an option to **publish** a repository into cloud. Publishing works for other local repositories, but it doesn't work for repositories created by PkgTemplates

**Comment by christopher-dG** [comment]
Okay, I'll try installing that and playing around with it.

**Comment by christopher-dG** [comment]
Okay, I got the same error. I'll play around with the remote setting later today.

**Comment by christopher-dG** [comment]
Turns out GitHub Desktop just isn't that smart... It assumes that if a remote exists, then the repo must already exist upstream. Which makes sense when people are using GitHub Desktop for everything, but that's far from universal.

**Comment by aminya** [comment]
@christopher-dG I created an issue in Github Desktop. I hope they respond. Otherwise, I will need to deactivate the Git option of PkgTemplate and create Git repo myself through Github Desktop.

**Comment by christopher-dG** [comment]
A simpler alternative would be to create a plugin whose `posthook` removed the remote.

**Comment by nickrobinson251** [comment]
> I have made the decision to wait until after this PR to do the interactive stuff.

Seems fair enough to me! Since this is package is such an entry-point for many users, and interactive is probably the easiest way to get started, it's a pretty important feature, so i wonder if we should hold of on tagging 0.7 until it is in?

File: docs/make.jl [hunk 15]
```diff
@@ -1,20 +1,21 @@
...
+using Documenter: Documenter, makedocs, deploydocs
+using PkgTemplates: PkgTemplates
 
 makedocs(;
     modules=[PkgTemplates],
-    format=Documenter.HTML(),
+    authors="Chris de Graaf, Invenia Technical Computing Corporation",
+    repo="https://github.com/invenia/PkgTemplates.jl/blob/{commit}{path}#L{line}",
+    sitename="PkgTemplates.jl",
+    format=Documenter.HTML(;
```
aminya:
```suggestion
    format=Documenter.HTML(;
            prettyurls = prettyurls = get(ENV, "CI", nothing) == "true",
```
When we have different markdawn files, this option helps to make separate html files in local build instead of folders.

aminya: Sory I meant this for the template not for PkgTemplate documentation itself

christopher-dG: Oh wow, I wish I had known this option existed before! Thanks.

aminya:
@christopher-dG Have you seen this:
https://github.com/julia-actions/example-workflows
It allows to use Github actions instead of .travis.yml, etc.

This can be a separate PR. Just wanted you to be aware of.

aminya:
@christopher-dG for pushing these ymls to Github's workflow some additional permissions are needed. I couldn't find them over the internet, but with Github desktop it grants those automatically.

aminya:
I think here is the solution: https://stackoverflow.com/questions/57921401/push-to-origin-from-github-action
Can you add these to the git config when Github actions are used?

christopher-dG:
Not sure if I'm interpreting this correctly, but wouldn't setting up a Git push be Documenter's responsibility?

aminya:
> 
> 
> Not sure if I'm interpreting this correctly, but wouldn't setting up a Git push be Documenter's responsibility? And even then, its pushes don't update workflow files. AFAIK no other GitHub Actions are trying to push.

The problem is that without Google Desktop, I cannot push any .yml file to Github's workflow folder. So probably if the user generates a yml file using this template, they can't push it. 
I think this should be solved by individual basis (they change their pushing command).

christopher-dG:
Ok, after some digging, I'm guessing that you use a personal access token to authenticate with GitHub instead of your password? Because this issue only seems to affect access tokens, and I've just tried with both password and SSH and had no problems.
You might want to look into using SSH keys if you want to avoid using your password for Git!

christopher-dG:
:smiley:

aminya:
Congratulation 🚀  @christopher-dG

PR #63, opened 2019-03-10
Add EUPL-1.2+ (EN) license

SaschaMann:
Add the [European Union Public Licence v1.2](https://joinup.ec.europa.eu/collection/eupl/eupl-text-11-12) to the templates.

---

Since this licence exists and is valid in 22 languages, and devs can choose their preferred one, I'd also like to add those versions. (mostly because I also need the German version in addition to the English one for some of my own projects and it'd make my life easier 😛)
However, it would blow up the interactive prompt, so I decided against it. If you have a suggestion on how to get around that issue, I'd implement it and submit another PR.

### DIFF
--- a/licenses/EUPL-1.2+
+++ b/licenses/EUPL-1.2+
@@ -189,4 +189,3 @@ The European Commission may update this Appendix to later versions of the above 
 a new version of the EUPL, as long as they provide the rights granted in Article 2 of this Licence and protect the 
 covered Source Code from exclusive appropriation. 
 All other changes or additions to this Appendix require the production of a new EUPL version. 
-
--- a/licenses/EUPL-1.2+
+++ b/licenses/EUPL-1.2+
@@ -1,0 +1,192 @@
+
+EUROPEAN UNION PUBLIC LICENCE v. 1.2 
+EUPL © the European Union 2007, 2016 
+
+This European Union Public Licence (the ‘EUPL’) applies to the Work (as defined below) which is provided under the 
+terms of this Licence. Any use of the Work, other than as authorised under this Licence is prohibited (to the extent such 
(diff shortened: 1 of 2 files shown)


licenses/EUPL-1.2+
```diff
@@ -0,0 +1,192 @@
...
+— GNU Lesser General Public Licence (LGPL) v. 2.1, v. 3 
+— Creative Commons Attribution-ShareAlike v. 3.0 Unported (CC BY-SA 3.0) for works other than software 
+— European Union Public Licence (EUPL) v. 1.1, v. 1.2 
+— Québec Free and Open-Source Licence — Reciprocity (LiLiQ-R) or Strong Reciprocity (LiLiQ-R+).
+
+The European Commission may update this Appendix to later versions of the above licences without producing 
+a new version of the EUPL, as long as they provide the rights granted in Article 2 of this Licence and protect the 
+covered Source Code from exclusive appropriation. 
+All other changes or additions to this Appendix require the production of a new EUPL version. 
+
```
christopher-dG: Could you drop this empty line please?

christopher-dG:
Hm, interesting question on how we can enable multiple languages. This would be easily accomplished with a plugin which took a language argument, but I don't recommend implementing anything until #61 is finished as many internal APIs are changing.

SaschaMann:
Removed the extra line.

I'll take another look at adding the other languages once #61 is merged then.

SaschaMann:
Forgot that I added (EN) in the full name. Removed it, it doesn't make sense as long as there's only one version in the package.

christopher-dG:
Great, thanks :slightly_smiling_face:

SaschaMann:
Thanks for the quick merge :)

PR #70, opened 2019-04-11
Revert "Use METADATA-compatible UUIDs"

tkf:
This partially reverts commit 8a1b65a34a43ee80ae868ec4c2775a716f9b8352 from #56.

see: https://discourse.julialang.org/t/22677/22

### DIFF
--- a/src/generate.jl
+++ b/src/generate.jl
@@ -18,11 +18,6 @@ function generate(
         # Create the directory with some boilerplate inside.
         Pkg.generate(pkg_dir)
 
-        # Replace the UUID with something that's compatible with METADATA.
-        project = joinpath(pkg_dir, "Project.toml")
-        uuid = string(Pkg.METADATA_compatible_uuid(pkg))
-        write(project, replace(read(project, String), r"uuid = .*" => "uuid = \"$uuid\""))
-
         if git
             # Initialize the repo.
             repo = LibGit2.init(pkg_dir)


christopher-dG:
I don't think I want to do this quite yet, there are still a fair number of people relying on METADATA (in fact, the Registrator switch-over isn't even done yet). And having this in there is not really hurting anything. So I'll let it sit for a while.

tkf:
I was thinking that this would make sense after the Registrator is online and hoping that this PR would get reviewed by around that time.  But apparently it complicated the review process.  Sorry about that.

> having this in there is not really hurting anything

I think there is a harm in using `METADATA_compatible_uuid`.  One of the great features of Pkg is that the package identity is not the name and different packages with the same name can coexist in one environment.  So, I can use arbitrary name for some internal packages without worrying about name crash with the ones in JuliaRegistries/General.  It would be nice to get this feature when you are using PkgTemplates.  Also, wouldn't `METADATA_compatible_uuid` be useless after the switch to General?

christopher-dG:
That's a good point about not allowing multiple packages with same name, thanks. 
I'll definitely look at merging this when everything registration-related is settled :slightly_smiling_face:.

tkf:
Sounds great. Thanks :+1:

christopher-dG:
Ok, I think by now we can safely merge this 👍

PR #73, opened 2019-05-14
Add CITATION.bib

matbesancon:
Since the Julia community is starting to agree on a standardized CITATION.bib file, it would be great to be able to have it from PkkTemplates

### DIFF
--- a/src/generate.jl
+++ b/src/generate.jl
@@ -216,7 +216,7 @@ function gen_readme(pkg_dir::AbstractString, t::Template)
             "\n",
         )
     end
-    if haskey(t.plugins, Citation) && p.plugins[Citation].readme_section
+    if haskey(t.plugins, Citation) && t.plugins[Citation].readme_section
         text *= "\n## Citing\n\nSee `CITATION.bib` for the relevant reference(s)\n"
     end
 
--- a/src/generate.jl
+++ b/src/generate.jl
@@ -216,7 +216,7 @@ function gen_readme(pkg_dir::AbstractString, t::Template)
             "\n",
         )
     end
-    if haskey(p.plugins, Citation) && p.plugins[Citation].readme_section
+    if haskey(t.plugins, Citation) && p.plugins[Citation].readme_section
         text *= "## Citing\n\nSee `CITATION.bib` for the relevant reference(s)\n"        
     end
 
--- a/src/generate.jl
+++ b/src/generate.jl
@@ -63,7 +63,6 @@ function generate(
             gen_require(pkg_dir, t),
             gen_readme(pkg_dir, t),
             gen_license(pkg_dir, t),
-            gen_citation(pkg_dir, t),
             vcat(map(p -> gen_plugin(p, t, pkg), values(t.plugins))...),
         )
 
@@ -216,12 +215,9 @@ function gen_readme(pkg_dir::AbstractString, t::Template)
             badges(t.plugins[plugin_type], t.user, pkg),
             "\n",
         )
-        if plugin_type <: Citation
-            c = t.plugins[plugin_type]
-            if c.readme_section
-                text *= "## Citing\n\nSee `CITATION.bib` for the relevant reference(s)\n"
-            end
-        end
+    end
+    if haskey(p.plugins, Citation) && p.plugins[Citation].readme_section
+        text *= "## Citing\n\nSee `CITATION.bib` for the relevant reference(s)\n"        
     end
 
 
--- a/src/generate.jl
+++ b/src/generate.jl
@@ -283,7 +283,7 @@ Create a CITATION.bib in `pkg_dir`.
 
 Returns an array of generated file/directory names.
 """
-function gen_license(pkg_dir::AbstractString, t::Template)
+function gen_citation(pkg_dir::AbstractString, t::Template)
     if !t.citation
         return String[]
     end
--- a/src/generate.jl
+++ b/src/generate.jl
@@ -63,6 +63,7 @@ function generate(
             gen_require(pkg_dir, t),
             gen_readme(pkg_dir, t),
             gen_license(pkg_dir, t),
+            gen_citation(pkg_dir, t),
             vcat(map(p -> gen_plugin(p, t, pkg), values(t.plugins))...),
         )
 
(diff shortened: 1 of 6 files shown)


christopher-dG:
I would advise adding this as a plugin instead of some special case. The package is in a bit of a weird state too, where the internals are completely redone in #61... I need to finish that to make it easier to contribute.

matbesancon:
so #61 will be blocking to switch to the plugin mechanism right? From the doc I wasn't sure this was supposed to be a plugin:
> Generic plugins are plugins that add any number of patterns to the generated package's .gitignore

Since this was not related to `.gitignore`, I was doubting on it being a plugin

christopher-dG:
Oh I suppose the docs could be clearer, a GenericPlugin is basically something that has the capability to create a single file in the package (i.e. CITATION.bib) and/or add entries to the gitignore. Thanks for bringing that up.

The PR is not required to make this a plugin, it's just that the plugin would be rewritten to be one line with the new internals. If you'd like to get this merged before then, feel free to implement as a `Citation` plugin (you can basically copy the `travisci.jl` code to suit your needs). 

Again, sorry contributing is a bit messy at the moment 🙃.

matbesancon:
this is indeed hairy, I think I managed to get something, still need to test it a bit

src/plugins/citation.jl
```diff
@@ -0,0 +1,43 @@
...
+            readme_section,
+        )
+    end
+end
+
+interactive(::Type{Citation}) = interactive(Citation; readme_section=false)
+
+function gen_plugin(p::Citation, t::Template, pkg_name::AbstractString)
+    pkg_dir = joinpath(t.dir, pkg_name)
+    text = "@misc{$pkg_name.jl,\n"
```
christopher-dG: This might be prettier with something like:

```julia
lines = [
    "@misc{$pkg_name.jl,"
    # ...
]
text = join(lines, "\n")
```

]

src/generate.jl
```diff
@@ -215,8 +216,15 @@ function gen_readme(pkg_dir::AbstractString, t::Template)
             badges(t.plugins[plugin_type], t.user, pkg),
             "\n",
         )
+        if plugin_type <: Citation
```
christopher-dG: This should be outside of (after) the loop, something like this:

```julia
if haskey(p.plugins, Citation) && p.plugins[Citation].readme_section
  text *= "\n## Citing\n\n ..."
end
```

src/generate.jl
```diff
@@ -63,6 +63,7 @@ function generate(
             gen_require(pkg_dir, t),
             gen_readme(pkg_dir, t),
             gen_license(pkg_dir, t),
+            gen_citation(pkg_dir, t),
```
christopher-dG: I think this can be deleted now.

src/generate.jl
```diff
@@ -216,6 +216,10 @@ function gen_readme(pkg_dir::AbstractString, t::Template)
             "\n",
         )
     end
+    if haskey(t.plugins, Citation) && p.plugins[Citation].readme_section
+        text *= "## Citing\n\nSee `CITATION.bib` for the relevant reference(s)\n"
```
christopher-dG:
```suggestion
        text *= "\n## Citing\n\nSee `CITATION.bib` for the relevant reference(s).\n"        
```

This will result in the following, I think:
```
[badge 1]
[badge 2]

## Citing

See `CITATION.bib` for asdf.
```

src/template.jl
```diff
@@ -24,6 +24,7 @@ create a template, you can use [`interactive_template`](@ref) instead.
 * `julia_version::VersionNumber=$VERSION`: Minimum allowed Julia version.
 * `ssh::Bool=false`: Whether or not to use SSH for the git remote. If `false` HTTPS will be used.
 * `manifest::Bool=false`: Whether or not to commit the `Manifest.toml`.
+* `citation::Bool=false`: Whether or not to create a `CITATION.bib` file at the top-level.
```
christopher-dG: Everything in this file can go now that Citation is a plugin

src/plugins/citation.jl
```diff
@@ -0,0 +1,43 @@
...
+            readme_section,
+        )
+    end
+end
+
+interactive(::Type{Citation}) = interactive(Citation; readme_section=false)
+
+function gen_plugin(p::Citation, t::Template, pkg_name::AbstractString)
+    pkg_dir = joinpath(t.dir, pkg_name)
+    text = """@misc{$pkg_name.jl,
```
christopher-dG: Let's format this as:
```jl
text = """
    @misc{$pkg_name.jl,
    \tauthor = ...
    \t...
    }
    """
```
I think we also need to get rid of the `\n` at the end of the lines.

src/plugins/citation.jl
```diff
@@ -0,0 +1,43 @@
...
+              \tauthor  = {{$(t.author)}},\n
```
christopher-dG: Are there supposed to be two curlies like {{this}} or just {one}?

src/plugins/citation.jl
```diff
@@ -0,0 +1,43 @@
...
+              \ttitle   = {{$(pkg_name).jl}},\n
```
christopher-dG:
```suggestion
              \ttitle   = {{$pkg_name.jl}},\n
```

Also, same question as `author`.

src/plugins/citation.jl
```diff
@@ -0,0 +1,43 @@
...
+              \turl     = {https://$(t.host)/$(t.user)/$(pkg_name).jl},\n
+              \tversion = {v0.0.1},\n
```
christopher-dG:
```suggestion
              \tversion = {v0.1.0},\n
```

test/plugins/citation.jl
```diff
@@ -0,0 +1,29 @@
...
+        @test p.dest == "CITATION.bib"
+        @test isempty(p.badges)
+        @test isempty(p.view)
+        @test !p.readme_section
+        p = Citation(; readme_section=true)
+        @test p.readme_section
+    end
+
+    @testset "File generation" begin
+        # Without a coverage plugin in the template, there should be no post-test step.
```
christopher-dG: This looks like it might be a copy-paste remnant

test/plugins/citation.jl
```diff
@@ -0,0 +1,29 @@
...
+        p = Citation()
+        @test gen_plugin(p, t, test_pkg) == ["CITATION.bib"]
+        @test isfile(joinpath(pkg_dir, "CITATION.bib"))
+        citation = read(joinpath(pkg_dir, "CITATION.bib"), String)
+
+        @test occursin("@misc", citation)
+        @test occursin("$(t.authors)", citation)
+        @test occursin("v0.0.1", citation)
```
christopher-dG:
```suggestion
        @test occursin("v0.1.0", citation)
```

matbesancon:
alright tests added and passing

src/generate.jl
matbesancon: fixed in last push

matbesancon:
@christopher-dG looks good on my side

src/generate.jl
```diff
@@ -216,6 +216,10 @@ function gen_readme(pkg_dir::AbstractString, t::Template)
             "\n",
         )
     end
+    if haskey(t.plugins, Citation) && t.plugins[Citation].readme_section
+        text *= "\n## Citing\n\nSee `CITATION.bib` for the relevant reference(s)\n"
```
christopher-dG:
```suggestion
        text *= "\n## Citing\n\nSee `CITATION.bib` for the relevant reference(s).\n"
```


Punctuation please 🙂

src/plugins/citation.jl
matbesancon: Two maintains the uppercase but your right it'n not necessary for authors

src/plugins/citation.jl
matbesancon: I find it easier on the eyes to have explicit parentheses when the variable name is long-ish, otherwise it can lead to ambiguities on what belongs to the interpolated expression: $(pkg_name).jl or $(pkg_name.jl)

src/generate.jl
matbesancon: fixed in last commit

christopher-dG:
Awesome, thanks :slightly_smiling_face:

PR #85, opened 2019-08-21
Use standard text (OSI)

Nosferican:
Currently many licenses are not machine detectable by GitHub/licensee. We should be using a standard text which it is machine detectable.


christopher-dG:
Is this actually necessary? Licensee says it ignores whitespace:

> If the license is an exact match to a known license. If we strip away whitespace and copyright notice, we might get lucky, and direct string comparison in Ruby is cheap.

I personally like having the file wrapped to 80 columns (the files come from GitHub's [choosealicense](https://choosealicense.com/licenses/), which is also [what Licensee uses](https://licensee.github.io/licensee/what-we-look-at/).

Thoughts?

Nosferican:
I believe in https://github.com/JuliaLang/Example.jl/pull/48 we decided to go with the one from Choose a license. For MIT, it is quite flexible and matches fine even with a header (i.e., first line `MIT License`). Would that be preferable? As long as it is machine detectable I don't have any strong opinions.

christopher-dG:
I think that is what is already done. Right now I'm just going through the licenses and making sure they're just copy-pasted in with no modifications other than templating for the year and author.

Somewhat related, in case you're more knowledgeable about this than I am:
If a license doesn't include any kind of "copyright <year> <owner>" line in it, should a line be added to the top to that effect, or should it be left out? For example the MIT license does explicitly include a place to put that information, but the Apache license does not.

Nosferican:
I like having a header with the license and SPDX code, but as you mention in some cases it breaks licensee's detection for some licenses. MIT is detected easily without or without a header, I believe some BSD are no longer detectable if one includes a header.

Nosferican:
Similarly, some licenses allow to specify the copyright holders while other do not. The best approach would be to have a `codemeta.json` with additional data, but no perfect solution based on license text alone.

christopher-dG:
Going to close this for now because I believe the license format in #61 will be sufficient.

PR #98, opened 2019-10-02
Test Julia 1.2 on Travis

ChrisRackauckas:
Julia 1.2 is the latest release and is quite standard, so therefore it would be nice to test on it as well :).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ os:
 julia:
   - 1.0
   - 1.1
+  - 1.2
   - nightly
 matrix:
   allow_failures:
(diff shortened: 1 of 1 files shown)


christopher-dG:
I'm 99.99999% sure that the failure is not due to the package, but bugs in the tests themselves. I'm going to close this since it becomes a non-issue when #61, where 1.2 testing is enabled and passing, is finally done. 

Sorry!

ChrisRackauckas:
It's all good. I just did this for an example video anyways 😄

ChrisRackauckas:
https://www.youtube.com/watch?v=QVmU29rCjaA 😸

PR #106, opened 2019-11-10
The default initial version number should be 1.0.0

DilumAluthge:
Currently, the default initial version number of a newly created package is `0.1.0`.

This pull request changes the default initial version number to `1.0.0`.

## Rationale

There has been a lot of confusion in the Julia community about the behavior of semver in pre-`1.0` versions.

I think that we should encourage people to just start numbering the versions of their packages at `1.0.0`.

Of course, it is extremely easy for a user to select their own initial version number simply by using the `ProjectFile` plugin, i.e.:
```julia
ProjectFile(; version=v"3.4.5")
```

So I don't think it is a huge burden for a user to change the initial version number.

But I think that we should encourage good default values, and I think `1.0.0` is a good default value for the initial version number.

### DIFF
--- a/docs/src/developer.md
+++ b/docs/src/developer.md
@@ -108,7 +108,7 @@ function hook(p::Documenter, t::Template, pkg_dir::AbstractString)
 
     make = render_file(p.make_jl, combined_view(p, t, pkg), tags(p))
     gen_file(joinpath(docs_dir, "make.jl"), make)
-    
+
     index = render_file(p.index_md, combined_view(p, t, pkg), tags(p))
     gen_file(joinpath(docs_dir, "src", "index.md"), index)
 
@@ -255,7 +255,7 @@ To see why, it might help to see the template file in its entirety:
 	author  = {<<&AUTHORS>>},
 	title   = {<<&PKG>>.jl},
 	url     = {<<&URL>>},
-	version = {v0.1.0},
+	version = {v1.0.0},
 	year    = {<<&YEAR>>},
 	month   = {<<&MONTH>>}
 }
--- a/src/plugins/project_file.jl
+++ b/src/plugins/project_file.jl
@@ -1,5 +1,5 @@
 """
-    ProjectFile(; version=v"0.1.0")
+    ProjectFile(; version=v"1.0.0")
 
 Creates a `Project.toml`.
 
@@ -7,7 +7,7 @@ Creates a `Project.toml`.
 - `version::VersionNumber`: The initial version of created packages.
 """
 @with_kw_noshow struct ProjectFile <: Plugin
-    version::VersionNumber = v"0.1.0"
+    version::VersionNumber = v"1.0.0"
 end
 
 # Other plugins like Tests will modify this file.
--- a/templates/CITATION.bib
+++ b/templates/CITATION.bib
@@ -2,7 +2,7 @@
 	author  = {<<&AUTHORS>>},
 	title   = {<<&PKG>>.jl},
 	url     = {<<&URL>>},
-	version = {v0.1.0},
+	version = {v1.0.0},
 	year    = {<<&YEAR>>},
 	month   = {<<&MONTH>>}
 }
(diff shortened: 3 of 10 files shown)


christopher-dG:
Strongly strongly strongly disagree. I get that 0.x SemVer is confusing for some people, but it exists for a reason (presumably to avoid hitting v10 in the first month of development).
I think there is a problem in the ecosystem here, but I don't think this is the solution.
I wouldn't block this if there's a consensus somewhere, but I think it needs a lot more discussion.

DilumAluthge:
The main problem here is that people start off with version `0.1.0` because that is the default value that they are given, then continue with `0.2.0`, `0.3.0`, `0.4.0`, etc.... and never end up releasing version `1.0.0`.

As a result, if you look at the General registry right now, approximately 85% of packages have a latest version of the form `0.y`. (https://discourse.julialang.org/t/how-can-we-encourage-julia-package-developers-to-release-version-1-0-0/29124)

I agree with you that `0.y` may be appropriate if your package is incredibly unstable and undergoing rapid changes in the public API.

But are we to believe that 85% of packages that are publicly registered in the General registry are incredibly unstable?

In my opinion, semver is not supposed to be used primarily in the `0.y.z` area. I think that semver should be used in a way where most packages have versions of the form `x.y.z` where `x` is nonzero. And the special case of `0.y.z` is supposed to be reserved for the very early rapid development stages of a package. It should not be the default stage for most packages.

One way or another, we need to get people to start releasing version `1.0` of their packages so that we can actually use semver properly.

I have suggested multiple approaches to this:

1. I made the Discourse thread to solicit concrete suggestions https://discourse.julialang.org/t/how-can-we-encourage-julia-package-developers-to-release-version-1-0-0/29124. The responses to the thread just illustrated the cultural problems, i.e. (1) people think of `1.0.0` as a magic version number that they can only use once their package is completely "finished", and thus they never end up releasing `1.0.0`, and (2) people do not understand the problems with `0.y` packages, for example, we have no guarantee of backwards compatibility from `0.y` to `0.(y+1`).

2. When I wrote the automerge bot, I originally had the bot post a comment if you were registering a package with version of the form `0.y`. The bot posted a comment saying something like "I see that you are registering a version of the form `0.y`. Please consider releasing version `1.0` if your public API is relatively stable." This comment was later removed because multiple people thought it made the automerge comment too lengthy and verbose.

3. I suggested that we disable automerging entirely for all releases of the form `0.y`, and only allow automerging releases where the version is `>= 1.0.0`. Then, people would be forced to release `1.0.0` in order to get the benefit of automerging. (You could still register `0.y`, but it would need to be manually merged.) Nobody liked this idea (except for @davidanthoff 🙂 ). I doubt that this idea will ever be implemented.

4. I have suggested that we set the default initial version number to `1.0.0` when generating packages. (This PR, and see also https://github.com/JuliaLang/Pkg.jl/pull/1497.) It sounds like you are not a fan, and I am guessing that https://github.com/JuliaLang/Pkg.jl/pull/1497 will get similar opposition.

Honestly I am running out of ideas here.

christopher-dG:
I understand your concerns and frustration, but encouraging **new** packages to be 1.0 does not solve the problem of **old** packages not releasing 1.0. 1.0 is for packages that have achieved stability (at least that's my interpretation of SemVer), and that requires iteration and experimentation.
It's a tough nut to crack.

DilumAluthge:
In my opinion, if other people are using your package, you should release 1.0. And personally, I think that if you are registering your package in the General registry, then you are doing so because you want other people to use your package.

From the horse's mouth:

> How do I know when to release 1.0.0?
> If your software is being used in production, it should probably already be 1.0.0. If you have a stable API on which users have come to depend, you should be 1.0.0. If you’re worrying a lot about backwards compatibility, you should probably already be 1.0.0.

DilumAluthge:
It is certainly very frustrating.

DilumAluthge:
My latest idea is that we only automerge non-breaking releases: https://github.com/JuliaRegistries/RegistryCI.jl/issues/103

christopher-dG:
From what I remember, General was explicitly meant to be the "anything goes" registry, so it's to be expected that plenty of packages there are not production ready (I've released a few of that nature myself :sweat_smile:). It hasn't really panned out that way though, since we don't have a "Cathedral" alternative for it to compare against.

PR #79, opened 2019-06-17
integrate OnlinePackage.jl

bramtayl:

PR #107, opened 2019-11-10
Add plugins for TagBot and CompatHelper

### LINKED ISSUE #105: Feature request: Auto-generate the GitHub Actions workflow file for CompatHelper.jl
Would this be easy or difficult?

Basically, we would need to generate the file `.github/workflows/CompatHelper.yml` with the following content.

(Presumably, instead of hard-coding `1.2.0`, we would need to replace that with whatever version of Julia the user chose when creating their `Template`.)

```yaml
name: CompatHelper

on:
  schedule:
    - cron: '00 * * * *'

jobs:
  build:
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        julia-version: [1.2.0]
        julia-arch: [x86]
        os: [ubuntu-latest]
    steps:
      - uses: julia-actions/setup-julia@latest
        with:
          version: ${{ matrix.julia-version }}
      - name: Pkg.add("CompatHelper")
        run: julia -e 'using Pkg; Pkg.add("CompatHelper")'
      - name: CompatHelper.main()
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        run: julia -e 'using CompatHelper; CompatHelper.main()'
```

christopher-dG:
Closes #105

TagBot is a default plugin due to its wide adoption, but I think I will wait until CompatHelper hits 1.0 before making it default.


crstnbr:
May I add that it would be great if these plugins were to support custom registries right from the start.

For example, they could take a `custom_registry` keyword that would trigger the supplementation of the workflows:

* Add `registry: user/CustomRegistry`  entry to `TagBot.yml`.
* Use `CompatHelper.main(; registries=...)` (see [here](https://github.com/crstnbr/TestPackage2.jl/blob/master/.github/workflows/CompatHelper.yml) for an example)

christopher-dG:
I'm not going to do the CompatHelper one for now because it wouldn't have a very nice API (`registries=[(name=..., uuid=..., url=...)]` I guess), but I just pushed the TagBot changes I was sitting on. 
The CompatHelper plugin does have a `file` option though, so you can just supply your own workflow file with any registries you need.

crstnbr:
> I'm not going to do the CompatHelper one for now because it wouldn't have a very nice API (`registries=[(name=..., uuid=..., url=...)]` I guess)

If we were to restrict ourselves to just a single custom repository I don't think the API would have to be like this. As I said above, we could use the same API as for TagBot: from "owner/registryrepo" we can generate the URL "http://github.com/owner/registryrepo/Registry.toml", where we find the name and uuid information of the custom registry.

christopher-dG:
Yeah I'm cool with that API. But since that requires HTTP, I'm going to wait on implementing it until we do more HTTP-related things (#79).

christopher-dG:
This is good to go, I'm just thinking about how I want to deal with https://github.com/JuliaRegistries/TagBot/issues/36 (the TagBot template will change if I do the underscored inputs).

DilumAluthge:
@christopher-dG Just FYI- I just registered CompatHelper version 1.0.0. It will be automerged within an hour.

Since the CompatHelper API is stable, how do you feel about it making it a default?

christopher-dG:
Yeah I'm cool with defaulting it, seems like it's been very well-received by the community.
Would love to see https://github.com/bcbi/CompatHelper.jl/issues/123 released first though, since that makes the templating much easier and prettier :slightly_smiling_face:

christopher-dG:
I'm gonna go ahead and merge this tonight (after I merge some stuff in TagBot). Custom registries are not supported in the default CompatHelper template, but you can bring your own if you need for now. Default support will come when CompatHelper has a nice config option for it.

PR #108, opened 2019-11-10
Allow failures for Windows on 1.0

christopher-dG:
FFTW, is required by something in ReferenceTests, and its build seems to either hang or just take a long time on Windows in Julia 1.0. 
Example: https://travis-ci.org/invenia/PkgTemplates.jl/jobs/609854008

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ julia:
 before_script:
   - git config --global user.name Tester
   - git config --global user.email te@st.er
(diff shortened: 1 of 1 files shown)


christopher-dG:
Welp, now the Linux 32-bit build just had the same error... I wonder if just replacing the default `script` but with `travis_wait` would be a better solution.

christopher-dG:
I don't really like this but it's better than an error badge :man_shrugging:

PR #119, opened 2019-12-06
Add missing Julia Github actions

abhishalya:
This adds missing new Julia Github actions.


abhishalya:
The failing build is due to the failing master. The changes made should be fine imo.

christopher-dG:
Sorry I never really got to this, I wasn't sure if I wanted it and there was another nearly identical PR so I kind of froze. I don't think that any one Julia formatter is widely-accepted or mature enough to be a default, so I'm going to close this for the time being.

PR #120, opened 2019-12-06
Adding missing Julia GitHub Actions

fungktt:


templates/github/workflows/ci.yml
```diff
@@ -38,6 +38,7 @@ jobs:
           arch: ${{ matrix.arch }}
       - uses: julia-actions/julia-buildpkg@latest
       - uses: julia-actions/julia-runtest@latest
+      - uses: julia-actions/julia-codeformat@latest
```
abhishalya: This should be added to every instance of ci.yml file in testing to make the tests pass. (Currently the master is failing so the build would still be red but the ci.yml's tests should pass after this change).

@christopher-dG Also, I'm not sure how we handle actions which are written is JS/TS (like this one). Since I think we have to manage their dependencies first (?)

PR #129, opened 2020-01-19
RFC: Add `inline` keyword argument to `TravisCI()` plugin

DilumAluthge:
It is still quite common that a package author needs to disable inlining in their Travis tests in order to get accurate code coverage information. It is inconvenient to have to do this manually.

This pull request adds a keyword argument named `inline` to the `TravisCI()` plugin. The default value is `true`.
- If `inline = true`, then nothing happens.
- If `inline = false`, then PkgTemplates adds a line to `.travis.yml` that sets `--inline=no` during the Travis `script`.

### DIFF
--- a/src/plugins/ci.jl
+++ b/src/plugins/ci.jl
@@ -100,6 +100,7 @@ end
         arm64=false,
         coverage=true,
         extra_versions=$DEFAULT_CI_VERSIONS,
+        inline=true,
     )
 
 Integrates your packages with [Travis CI](https://travis-ci.com).
@@ -115,6 +116,7 @@ Integrates your packages with [Travis CI](https://travis-ci.com).
 - `coverage::Bool`: Whether or not to publish code coverage.
   Another code coverage plugin such as [`Codecov`](@ref) must also be included.
 $EXTRA_VERSIONS_DOC
+- `inline::Bool`: Whether or not inlining should be enabled during the tests. Default is `true`. Setting this to `false` may improve the accuracy of code coverage.
 """
 @with_kw_noshow struct TravisCI <: BasicPlugin
     file::String = default_file("travis.yml")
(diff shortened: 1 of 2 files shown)


christopher-dG:
I think this could apply to AppVeyor too, right?

src/plugins/ci.jl
```diff
@@ -115,6 +116,7 @@ Integrates your packages with [Travis CI](https://travis-ci.com).
 - `coverage::Bool`: Whether or not to publish code coverage.
   Another code coverage plugin such as [`Codecov`](@ref) must also be included.
 $EXTRA_VERSIONS_DOC
+- `inline::Bool`: Whether or not inlining should be enabled during the tests. Default is `true`. Setting this to `false` may improve the accuracy of code coverage.
```
christopher-dG:
```suggestion
- `inline::Bool`: Whether or not inlining should be enabled during the tests.
  Setting this to `false` may improve the accuracy of code coverage.
```

templates/travis.yml
```diff
@@ -34,6 +34,9 @@ jobs:
       julia: {{{E_JULIA}}}
       {{/E_JULIA}}
 {{/EXCLUDES}}
+{{#DISABLE_INLINING}}
+script: julia --code-coverage --inline=no -e 'using Pkg; Pkg.build(); Pkg.test(coverage=true)'
```
christopher-dG:
```suggestion
script: julia --code-coverage --inline=no -e 'using Pkg; Pkg.build(); Pkg.test(; coverage=true)'
```

i like semicolons

src/plugins/ci.jl
```diff
@@ -166,6 +169,7 @@ function view(p::TravisCI, t::Template, pkg::AbstractString)
         "USER" => t.user,
         "VERSION" => format_version(t.julia),
         "VERSIONS" => versions,
+        "DISABLE_INLINING" => !p.inline,
```
christopher-dG: sort alphabetically pls :slightly_smiling_face:

DilumAluthge:
I completely forgot about this, and I no longer need it.

DilumAluthge:
We can revisit it in the future, or someone else can pick up where I left off.

PR #131, opened 2020-01-23
Use literal block style for multi-line YAML

omus:
Allows for Julia code to not require the use of semi-colon which is matches how most users will write Julia code. An additional benefit is that it allows for use of single line comments.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ julia:
 before_script:
   - git config --global user.name Tester
   - git config --global user.email te@st.er
-script: travis_wait julia --project -e 'using Pkg; Pkg.test(; coverage=true)'
+script: travis_wait julia --project -e 'using Pkg; Pkg.test(coverage=true)'
 matrix:
   fast_finish: true
   allow_failures:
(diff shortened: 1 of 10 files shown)


christopher-dG:
Thanks for the bump, I'll get to this on the weekend.

omus:
> Also, the test files will need to be updated (you can include("test/runtests.jl") and accept the prompts to update the files).

It would be good to document this in a CONTRIBUTING.md file

.travis.yml
```diff
@@ -12,7 +12,7 @@ julia:
 before_script:
   - git config --global user.name Tester
   - git config --global user.email te@st.er
-script: travis_wait julia --project -e 'using Pkg; Pkg.test(; coverage=true)'
+script: travis_wait julia --project -e 'using Pkg; Pkg.test(coverage=true)'
```
christopher-dG: I generally like using semicolons for all keyword arguments but these changes are fine with me.

omus: I also tend to use semicolons. In this particular case it seems worthwhile to be a bit more terse

PR #134, opened 2020-02-10
Merge `matrix:` and `jobs:` sections in .travis.yml

### LINKED ISSUE #132: Stable version output does not produce docs on travis
On the released version of `PkgTemplates.jl` (v0.6.3), the `.travis.yml` file contains both a `matrix` section and a `jobs` section.  

These are aliases, and because both exist, one overwrites the other (in this case, `matrix` overwrites `jobs`.

Example: https://travis-ci.com/kmsquire/ArgParse2.jl/jobs/281357698/config 
(The error is in the middle of the page.)

Because of this configuration issue, documentation generation is not run.  It can be fixed by merging the `matrix` and `jobs` sections.

This has already been done on `master`, but since that is not ready to be released yet, I'm wondering if there is interest in updating the latest tagged version and releasing v0.6.4, so that people using the released version don't run into this issue?

(I don't mind working on this--it should be a simple fix--although I don't think there's currently a target branch available for a pull request targeting v0.6.3/stable.)

If master is about to be released or there's no interest in releasing a v0.6.4, please close this.

kmsquire:
* These are aliases in travis, and defining them twice causes one to
  overwrite the other.  In this case, `matrix` was overwriting `jobs`,
  which caused `Documentation` not to run.

Fixes #132

Note: this targets the current release branch, not master.  It's already been fixed on master, so this is meant as a stop-gap fix until master is release.

### DIFF
--- a/defaults/travis.yml
+++ b/defaults/travis.yml
@@ -6,10 +6,6 @@ os:
 julia:
   - {{VERSION}}
   - nightly
-matrix:
-  allow_failures:
-    - julia: nightly
-  fast_finish: true
 notifications:
   email: false
 {{#COVERAGE}}
@@ -21,8 +17,11 @@ after_success:
   - julia -e 'using Pkg; Pkg.add("Coverage"); using Coverage; Coveralls.submit(process_folder())'
 {{/COVERALLS}}
 {{/COVERAGE}}
-{{#DOCUMENTER}}
 jobs:
+  allow_failures:
+    - julia: nightly
+  fast_finish: true
+{{#DOCUMENTER}}
   include:
     - stage: Documentation
       julia: {{VERSION}}


defaults/travis.yml
```diff
@@ -23,6 +19,9 @@ after_success:
 {{/COVERAGE}}
 {{#DOCUMENTER}}
```
christopher-dG: This `{{#DOCUMENTER}}` tag needs to move...

defaults/travis.yml
```diff
@@ -23,6 +19,9 @@ after_success:
...
 jobs:
+  allow_failures:
+    - julia: nightly
+  fast_finish: true
   include:
```
christopher-dG: ...to right above this line.

christopher-dG:
thanks :smile: I'll put in the release in a bit.

PR #139, opened 2020-02-14
Test documentation on latest Julia release

omus:
I just noticed that PkgTemplates wasn't included in the docs/Project.toml which is required in Julia 1.3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ matrix:
     - arch: x86
       julia: 1.0
     - stage: Documentation
-      julia: 1.0
+      julia: 1
       script:
         - git config --global user.name name
         - git config --global user.email email
--- a/docs/Manifest.toml
+++ b/docs/Manifest.toml
@@ -3,6 +3,16 @@
 [[Base64]]
 uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
 
+[[DataAPI]]
+git-tree-sha1 = "674b67f344687a88310213ddfa8a2b3c76cc4252"
+uuid = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
+version = "1.1.0"
+
+[[DataValueInterfaces]]
+git-tree-sha1 = "bfc1187b79289637fa0ef6d4436ebdfe6905cbd6"
+uuid = "e2d170a0-9d28-54be-80f0-106bbe20a464"
+version = "1.0.0"
+
 [[Dates]]
 deps = ["Printf"]
 uuid = "ade2ca70-3891-5945-98fb-dc099432e06a"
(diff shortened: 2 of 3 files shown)


christopher-dG:
Is there somewhere other than [the Documenter docs](https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#docs/Project.toml-1) that I can get info on this change? Just want to make sure I can see the whole picture.

omus:
This is actually a Julia 1.3 requirement rather than a Documenter requirement. I'm not sure the PR it's from but it would be from the [Pkg.jl](https://github.com/JuliaLang/Pkg.jl) repo.
```bash
$ git checkout ab9949a
$ julia-1.3 --project=docs/ docs/make.jl
ERROR: LoadError: ArgumentError: Package PkgTemplates not found in current path:
- Run `import Pkg; Pkg.add("PkgTemplates")` to install the PkgTemplates package.

Stacktrace:
 [1] require(::Module, ::Symbol) at ./loading.jl:887
 [2] include at ./boot.jl:328 [inlined]
 [3] include_relative(::Module, ::String) at ./loading.jl:1105
 [4] include(::Module, ::String) at ./Base.jl:31
 [5] exec_options(::Base.JLOptions) at ./client.jl:287
 [6] _start() at ./client.jl:460
in expression starting at /Users/omus/.julia/dev/PkgTemplates/docs/make.jl:2
```

omus:
Rebased changes.

christopher-dG:
Seems good to me. Is this (adding the package to its own `docs/{Project,Manifest}.toml`) what we should be doing for generated packages from now on too? Right now the documentation is built on 1.0 so it's not necessary AFAIK. Does the `docs/Manifest.toml` need to be updated whenever a new dependency is added to the package being documented?

nickrobinson251:
> Is this (adding the package to its own docs/{Project,Manifest}.toml) what we should be doing for generated packages from now on too?

Yes; `docs/` should have `docs/Project.toml` including the Package itself as a dependency to be able to generate the API docs. In particular, the [default `docs/make.jl` file](https://github.com/invenia/PkgTemplates.jl/blob/master/templates/docs/make.jl#L1) depends on the Package, so needs to be an explicit dependency.

> the Documenter plugin should probably be updated too.

Yep. Using the `Documenter` plugin, then running `julia --project=docs/ docs/make.jl` gives
```julia
ERROR: LoadError: ArgumentError: Package Foo not found in current path:
- Run `import Pkg; Pkg.add("Foo")` to install the Foo package.
```
(I don't think this ever worked)

omus:
> Right now the documentation is built on 1.0 so it's not necessary AFAIK. Does the `docs/Manifest.toml` need to be updated whenever a new dependency is added to the package being documented?

Unfortunately using a `docs/Manifest.toml` does mean it needs to be periodically updated. I think the CompatHelper GitHub action may assist with this. I think we can leave the default as having the CI do `Pkg.develop(PackageSpec(path=pwd()))` for documentation. The main advantage of having a `docs/Manifest.toml` is it ensures that when building documentation locally you're using the local project and not the latest release.

christopher-dG:
Alright I'll merge this and open a new issue to mention that we should probably update the Documenter plugin.

PR #140, opened 2020-02-14
Switch to using GitLab CI extends

omus:
Use GitLab CI's [`extends`](https://docs.gitlab.com/ee/ci/yaml/#extends) instead of YAML anchors. The `extends` is a bit more flexible and the syntax is a bit cleaner.

Note this change will conflict with #131.

### DIFF
--- a/templates/gitlab-ci.yml
+++ b/templates/gitlab-ci.yml
@@ -1,30 +1,30 @@
-.definitions:
-  script: &script
-    script:
-      - |
-        julia --project=@. -e '
-          using Pkg
-          Pkg.build()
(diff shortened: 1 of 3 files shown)


christopher-dG:
Needs a rebase but looks good to me, provided you've tested that this actually works properly with GitLab!

omus:
Rebased and I've validated that the change works with GitLab CI. I'll note that [`extends`](https://docs.gitlab.com/ee/ci/yaml/#extends) requires GitLab 11.3 which we may want to mention somewhere. Current version is 12.8

PR #146, opened 2020-03-15
add env for github actions docs building

mileslucas:
This adds an environemnt variable to the github actions ci documenter build. This makes sure that if a user is trying to plot something you avoid getting [errors like this](https://github.com/JuliaPlots/Plots.jl/issues/1182).


christopher-dG:
I don't really think this is general enough to go in here, we tend to not include anything package-specific.

mileslucas:
👍🏻

christopher-dG:
I think there's probably a case for some kind of general `env` option for the various CI plugins that would cover these situations. Or you can just use a custom template file.

PR #153, opened 2020-04-15
default authors: add "and contributors" (close #152)

### LINKED ISSUE #152: LICENSE files: add "and contributors" to default copyright
It's a best practice to include "and contributors" to the copyright statement in a LICENSE file since the copyright of a project does not solely belong to the original author unless the project has a copyright assignment agreement (which is not legal in some countries). I have been manually adding "and contributors" to the copyright line of the LICENSE file whenever I use PkgTemplates to generate a package, but it would arguably be better if that was the default.

StefanKarpinski:
closes #152 

Also make "contributors" the default when the user name isn't known.

### DIFF
--- a/src/template.jl
+++ b/src/template.jl
@@ -13,9 +13,10 @@ default_plugins() = [
 
 function default_authors()
     name = LibGit2.getconfig("user.name", "")
-    isempty(name) && return ""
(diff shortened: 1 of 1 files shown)


christopher-dG:
Seems good 👍

PR #157, opened 2020-04-17
Fix some GitLab URL issues

### LINKED ISSUE #155: GitLabCI badges don't recognise hosting service
I just created a module with GitLabCI using interactive_template() on a custom server (not gitlab.com).

I recognised that the badges (build and coverage) that were added automatically to the Readme didn't work. The error was that gitlab.com was used for the links and the custom server.

Original:
```
[![Coverage](https://gitlab.com/name/example.jl/badges/master/coverage.svg)](https://gitlab.com/name/example.jl/commits/master)
```
Fixed:
```
[![Coverage](https://customserver.com/name/example.jl/badges/master/coverage.svg)](https://customserver.com/name/example.jl/commits/master)
```

### LINKED ISSUE #156: GitLabCI "Build Status" has wrong path
The default link for the build status badge for GitLabCI in the readme directs to `.../badges/master/build.svg` while it should direct to `.../badges/master/pipeline.svg `

Original:
```
[![Build Status](https://gitlab.com/name/example.jl/badges/master/build.svg)](https://gitlab.com/name/example.jl/pipelines)
```
Fixed:
```
[![Build Status](https://gitlab.com/name/example.jl/badges/master/pipeline.svg)](https://gitlab.com/name/example.jl/pipelines)
```

[Source](https://docs.gitlab.com/ee/ci/pipelines/settings.html#pipeline-status-badge)

Thanks by the way for this super useful package!

christopher-dG:
Closes #156 
Closes #155 

Thanks for the reports @judober! Always good to get feedback on the less-frequently used parts like GitLab support. Could you test this out and confirm that it does what you expect! (you can use `] add PkgTemplates#cdg/gitlab-urls` to get this branch with Pkg)


judober:
Well, I cannot use it as before because `interactive_template()` apparently has changed.
I now get:
```
ERROR: ArgumentError: Git: Git hosting service username is required, set one with keyword `user="<username>"`
```
Trying `t = interactive_template(user="name")` results in 
```
ERROR: MethodError: no method matching interactive_template(; user="name")
```
As interactive_template is apparently deprecated, I also tried the suggested `t = Template(; interactive = true, user="name")` which gives me:
```
┌ Warning: Unrecognized keywords were supplied, see the documentation for help
│   kwargs =
│    Dict{Symbol,Any} with 1 entry:
│      :interactive => true
└ @ PkgTemplates ~/.julia/packages/PkgTemplates/zz7bJ/src/template.jl:100
```

christopher-dG:
Oh right, I just kind of assume people use master branch now. Oops! Can you use the `Template(...)` constructor instead?

judober:
Ok, it it fixed thanks!

I'm just wondering if I can get the pages for the documentation to work too...
I think the path there is also specific for gitlab.io but I'm not sure what would be the correct version.

Btw: I got this during compilation of PkgTemplates:
```
WARNING: using PkgTemplates.Documenter in module Main conflicts with an existing identifier.
```

christopher-dG:
Yeah there's a TODO to deal with custom GitLab Pages URLs, but because the URL can technically be anything I think we'd have to add a new field to `Documenter` which feels clunky right now, so I'm gonna wait and think about it some more.

And I've realized that currently the `Documenter` plugin and `Documenter` package can't coexist in the same Julia session, but I don't think I see that as too much of a problem because the situation where as user creates a new package and builds documentation in the same session seems pretty rare. If I get a bunch of user complaints about it then I'd probably change it to `Docs`.

PR #165, opened 2020-05-07
Update README.md

joergbuchwald:
fix function call


christopher-dG:
It's not incorrect, Template instances are callable

joergbuchwald:
But I get the following error:
ERROR: MethodError: objects of type Template are not callable

christopher-dG:
Use master branch 😊

joergbuchwald:
Ah okay, sorry. However, I think it is a good idea to point to both variants.

PR #169, opened 2020-05-12
Update the CompatHelper template

DilumAluthge:
CompatHelper now supports SSH deploy keys. This pull request updates the CompatHelper template to support this new feature.

---

Just a note: You'll notice that the new template has a line that looks like this:
```yaml
COMPATHELPER_PRIV: ${{ secrets.COMPATHELPER_PRIV }}
```

If the user does not want or need to use SSH deploy keys, they don't need to set the `COMPATHELPER_PRIV` secret. And if the `COMPATHELPER_PRIV` secret is not set, CompatHelper will silently ignore it.

So we can include this line in the template unconditionally. If the `COMPATHELPER_PRIV` secret is set, CompatHelper will automatically detect it and use it. Otherwise, it will be ignored.

cc: @christopher-dG

### DIFF
--- a/templates/github/workflows/CompatHelper.yml
+++ b/templates/github/workflows/CompatHelper.yml
@@ -1,14 +1,17 @@
 name: CompatHelper
+
 on:
   schedule:
     - cron: <<&CRON>>
+
 jobs:
-  build:
+  CompatHelper:
     runs-on: ubuntu-latest
     steps:
       - name: Pkg.add("CompatHelper")
         run: julia -e 'using Pkg; Pkg.add("CompatHelper")'
       - name: CompatHelper.main()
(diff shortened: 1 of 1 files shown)


christopher-dG:
I'd personally prefer if the env line wasn't included for an unset variable.

Also, it would be nice for the name of the secret to be customizable so that deploy keys for Documenter/TagBot/etc. could be reused. You could look at the TagBot code to replicate that perhaps.

christopher-dG:
Looking back on this, it's really dumb for me to try to tell you what the config file for **your** tool should look like. @DilumAluthge if you run the tests locally, you should be able to easily update the reference files and CI will pass.

DilumAluthge:
No worries man! I completely forgot about this PR. Anyway, I am closing it in favor of #205

PR #171, opened 2020-05-25
Update version to 0.7.0

christopher-dG:
I think it's finally ready...


christopher-dG:
alright we're doing it 🤷 
if something is broken, we'll just make a new release instead of waiting 8 months!

PR #172, opened 2020-05-26
Make the .jl remote suffix optional

christopher-dG:
Had a nice conversation with @aminya who is looking at PkgTemplates for non-Julia stuff. All the Julia-specific plugins can be disabled easily, but Git only requires this small change to be language-agnostic.


aminya:
Thank you for making this ❤️.

PR #173, opened 2020-05-26
Cache artifacts in Travis, AppVeyor, GitHubActions, and Cirrus

### LINKED ISSUE #166: Cache artifacts in CI
It may be a good idea to setup CI scripts to automatically cache artifacts: https://discourse.julialang.org/t/recommendation-cache-julia-artifacts-in-ci-services/35484

christopher-dG:
Closes #166
I've tested Travis, AppVeyor, and GHA, but I couldn't get Cirrus to pick up on my repo for some reason. According to the docs it should work.
GitLab doesn't allow caching files outside of the repository directory, so caching can't be used there. DroneCI's caching looks a bit more complicated so maybe I'll look at it another day, it's not used frequently anyways.

### DIFF
--- a/templates/appveyor.yml
+++ b/templates/appveyor.yml
@@ -8,6 +8,8 @@ platform:
 {{#PLATFORMS}}
   - {{{.}}}
 {{/PLATFORMS}}
+cache:
+  - '%USERPROFILE%\.julia\artifacts'
 {{#HAS_ALLOW_FAILURES}}
 matrix:
   allow_failures:
--- a/templates/cirrus.yml
+++ b/templates/cirrus.yml
@@ -2,6 +2,8 @@ freebsd_instance:
   image: {{{IMAGE}}}
 task:
   name: FreeBSD
+  artifacts_cache:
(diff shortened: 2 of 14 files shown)


giordano:
Awesome, thank you so much!

Cirrus is being very unreliable lately, many packages moved to Travis for FreeBSD testing.  As far as I know, Cloud Drone doesn't support caching, there are plugins for self-hosted instances.

christopher-dG:
How does FreeBSD on Travis work? And good to know that the Drone stuff will only work on self-hosted, so not much use implementing it here.

giordano:
> How does FreeBSD on Travis work?

Pretty good, just add
```yml
  - freebsd
```
to the list of operating systems, that's it

christopher-dG:
It seems that it's not mentioned in the Travis docs anywhere...

giordano:
I think it's in beta or something like that, but Alex shared the news on Slack :slightly_smiling_face:

christopher-dG:
Ah, cool. I'll wait until it's GA before I add the option here.

PR #177, opened 2020-05-28
Show the value that will be used if not customized

rapus95:


christopher-dG:
I think I like this idea, because seeing the default value definitely gives a nice hint as to what the option actually does before you commit to customizing it.

I wonder if with this change we could change the input parsing to turn empty strings into `nothing`, instead of the default value. I think less people would "accidentally" choose to customize a value and then enter an empty string to get the default if they knew what the default value was right away.

Could you also make the same change in `interactive(T::Type)`?

src/template.jl
```diff
@@ -186,7 +186,10 @@ function interactive(::Type{Template}; kwargs...)
 
     return try
         println("Template keywords to customize:")
-        menu = MultiSelectMenu(map(string, customizable); pagesize=length(customizable))
+        menopts = map(customizable) do kw
+           string(kw, " (", defaultkw(Template, kw), ")") 
+        end
+        menu = MultiSelectMenu(menopts; pagesize=length(customizable))
```
christopher-dG, line 192:
```suggestion
        opts = map(k -> "$k ($(repr(defaultkw(Template, k))))" , customizable)
        menu = MultiSelectMenu(opts; pagesize=length(customizable))
```

christopher-dG:
I extended this PR a bit and came up with this:

```jl
julia> Template(; interactive=true)
Template keywords to customize:
[press: d=done, a=all, n=none]
   [ ] user ("christopher-dG")
   [ ] authors ("Chris de Graaf <me@cdg.dev> and contributors")
   [X] dir ("~/.local/share/julia/dev")
   [ ] host ("github.com")
   [ ] julia (v"1.0.0")
 > [X] plugins (PkgTemplates.Plugin[])
Enter value for 'dir' (String): ~/code
Select plugins:
[press: d=done, a=all, n=none]
   [ ] CompatHelper
   [ ] ProjectFile
   [ ] SrcDir
   [ ] Git
   [ ] License
   [ ] Readme
   [ ] Tests
   [X] TagBot
   [ ] AppVeyor
   [ ] CirrusCI
   [ ] Citation
 > [X] Codecov
   [ ] Coveralls
   [ ] Develop
   [ ] Documenter
   [ ] DroneCI
   [ ] GitHubActions
   [ ] GitLabCI
   [X] TravisCI
TagBot keywords to customize:
[press: d=done, a=all, n=none]
 > [ ] branches (nothing)
   [ ] changelog (nothing)
   [ ] changelog_ignore (nothing)
   [ ] cron ("0 0 * * *")
   [ ] destination ("TagBot.yml")
   [ ] dispatch (nothing)
   [ ] dispatch_delay (nothing)
   [ ] file ("~/code/PkgTemplates/templates/github/workflows/TagBot.yml")
   [ ] gpg (nothing)
   [ ] gpg_password (nothing)
   [ ] registry (nothing)
   [ ] ssh (nothing)
   [ ] ssh_password (nothing)
   [ ] token (Secret("GITHUB_TOKEN"))
Codecov keywords to customize:
[press: d=done, a=all, n=none]
 > [ ] file (nothing)
   [ ] None (No customization)
TravisCI keywords to customize:
[press: d=done, a=all, n=none]
 > [ ] arm64 (false)
   [ ] coverage (true)
   [ ] extra_versions (["1.0", "1.4", "nightly"])
   [ ] file ("~/code/PkgTemplates/templates/travis.yml")
   [ ] linux (true)
   [ ] osx (true)
   [ ] windows (true)
   [ ] x64 (true)
   [ ] x86 (false)
```

gdalle:
@oxinabox asking for a review because this might seem slightly overwhelming for beginning users (it adds a lot of info to the interactive menus), but at the same time I think the addition is welcome.

oxinabox (approved):
I think this is a net improvement

PR #181, opened 2020-06-03
Update Github actions CI

ericphanson:
The first commit uses the codecov action (https://github.com/codecov/codecov-action) which @SaschaMann recommended to me (pointing to https://github.com/uva-bi-sdad/COVID_Tracking_Clean/blob/master/.github/workflows/ci.yml#L64-L70) when I was having issues with codecov mixing up which branch was which. One nice thing about it is that a token is not required for public repos.

The second commit is a bit more opinionated (and I understand if you don't want to include it) and runs CI only on pushes or PRs to the master branch, and ignores changes to the license, readme, and CompatHelper and Tagbot actions. I end up almost always wanting this in my repos, especially since github actions don't autocancel (so push a few commits and then make a PR, and CI doesn't run on the PR for awhile since it has to process through the commits). I imagine it might be better if ignoring CompatHelper and Tagbot are gated on those plugins being used, but I wasn't sure how to do that and ignoring the paths doesn't really hurt anyway.

### DIFF
--- a/templates/github/workflows/ci.yml
+++ b/templates/github/workflows/ci.yml
@@ -52,10 +52,7 @@ jobs:
       - uses: julia-actions/julia-processcoverage@v1
       - uses: codecov/codecov-action@v1
         with:
-          file: ./lcov.info
-          flags: unittests
-          name: codecov-umbrella
-          fail_ci_if_error: false
+          file: lcov.info
       <</HAS_CODECOV>>
       <<#HAS_COVERALLS>>
       - uses: julia-actions/julia-uploadcoveralls@latest
--- a/templates/github/workflows/ci.yml
+++ b/templates/github/workflows/ci.yml
@@ -49,9 +49,13 @@ jobs:
       - uses: julia-actions/julia-buildpkg@latest
       - uses: julia-actions/julia-runtest@latest
       <<#HAS_CODECOV>>
-      - uses: julia-actions/julia-uploadcodecov@latest
-        env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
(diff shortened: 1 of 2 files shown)


SaschaMann:
More context for the first commit: https://github.com/julia-actions/julia-uploadcodecov/issues/1#issuecomment-608043596

templates/github/workflows/ci.yml
```diff
@@ -49,9 +63,13 @@ jobs:
       - uses: julia-actions/julia-buildpkg@latest
       - uses: julia-actions/julia-runtest@latest
       <<#HAS_CODECOV>>
-      - uses: julia-actions/julia-uploadcodecov@latest
-        env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+      - uses: julia-actions/julia-processcoverage@v1
+      - uses: codecov/codecov-action@v1
+        with:
+          file: ./lcov.info
```
christopher-dG, line 69: I think the `./` can probably be dropped

templates/github/workflows/ci.yml
```diff
@@ -49,9 +63,13 @@ jobs:
...
+          flags: unittests
+          name: codecov-umbrella
+          fail_ci_if_error: false
```
christopher-dG, line 72: I think these can be removed. They're all optional, the fail if error one is already false by default, and `unittests` isn't guaranteed to be a good description of the tests --- they might be integration tests etc. too!

ericphanson:
Ok, sounds good. Just in case, I tried out the change in parameters on my repo and it looks fine: https://github.com/ericphanson/GuessworkQuantumSideInfo.jl/pull/5/checks?check_run_id=735280172#step:7:16.

PR #188, opened 2020-06-09
Order Documenter and CI consistently

goretkin:


christopher-dG:
Thanks, good catch! I will deal with the test failure myself, there is something unrelated to this PR in addition to one new failure.

PR #198, opened 2020-06-30
Add default constructor for Documenter

### LINKED ISSUE #197: repr(Documentor) not evalable
I was following instructions on saving via using `show`
and then i `include`ed the file i created and i got an error because Documentor doesn't define the default constructor,
and thus can't be constructed with its positional arguments

It made for me:
```julia
Documenter{GitHubActions}(String[], Dict{Symbol,Any}(), PkgTemplates.github_pages_url, "/Users/oxinabox/.julia/packages/PkgTemplates/7Erns/templates/docs/make.jl", "/Users/oxinabox/.julia/packages/PkgTemplates/7Erns/templates/docs/src/index.md"),
```

trying to run that gives:
```
ERROR: MethodError: no method matching Documenter{GitHubActions}(::Array{String,1}, ::Dict{Symbol,Any}, ::typeof(PkgTemplates.github_pages_url), ::String, ::String)
```

christopher-dG:
Fixes #197, I think

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "PkgTemplates"
 uuid = "14b8a8f1-9102-5b29-a752-f990bacb7fe1"
 authors = ["Chris de Graaf", "Invenia Technical Computing Corporation"]
-version = "0.7.5"
+version = "0.7.6"
 
 [deps]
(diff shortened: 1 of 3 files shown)


christopher-dG:
@oxinabox feel free to try this out and merge if it fixes your issue.

oxinabox:
Yep works

PR #208, opened 2020-08-02
Move from Travis to GitHub Actions, fix Windows Git issue

### LINKED ISSUE #211: Fresh packages on Windows contain useless directory/symlink _git2_*
When I generate a new package on Windows, the folder contains a broken symlink with the name `_git2_*`, e.g. `_git2_a11248: broken symbolic link to testing`

Likely due to this issue: https://github.com/libgit2/libgit2/issues/5147

Later versions of libgit2 don't contain that issue anymore, could the version that PkgTemplates uses be updated to resolve this?

christopher-dG:
This is not ready yet, the reference tests fail on Windows ([link](https://github.com/invenia/PkgTemplates.jl/runs/936659973?check_suite_focus=true)) for some reason. It says it's trying to check file with names like `C:\Users\RUNNER~1\AppData\Local\Temp\jl_HADs2B\dev\Basic\_git2_a06688`, I have no idea where that basename is coming from.

GHA runs in about 10 minutes, whereas Travis takes about 18 (and it gets worse in the middle of  a work day).

Closes #211


SaschaMann:
> I have no idea where that basename is coming from.

I think it's due to #211

christopher-dG:
So this is working now and #211 is fixed, but the failed nightly builds are marking this PR as failing despite `continue-on-error: true`. @SaschaMann do you know if there's a way for me to get the status passing with nightly failures?

SaschaMann:
@christopher-dG I'm not entirely sure what's going wrong, I would expect it to work. My guess would be that `${{ matrix.version == 'nightly' }}` doesn't evaluate to `true` for some reason. Maybe it's `"true"`? Could you enable debug logs by setting the secret `ACTIONS_STEP_DEBUG` to `true` in this repo and then rerunning the build? I think you may have to paste the relevant logs for the step because I'm not sure if one can read debug logs without access to the repo.

christopher-dG:
[Here's a log dump of a failing nightly run](https://pastebin.com/2kV6Fz4Z) but I don't think there's anything relevant in there unfortunately.
One thing I did notice is that if I remove the `continue-on-error` line, then all the jobs get canceled as soon as one nightly job fails, whereas right now you can see the 1.0 and 1.x chugging away.

DilumAluthge:
Maybe try this:

```
continue-on-error: matrix.version == 'nightly'
```

Or maybe this:

```
continue-on-error: ${{ matrix.version }} == 'nightly'
```

-- 

Instead of what you currently have, which is this:
```
continue-on-error: ${{ matrix.version == 'nightly' }}
```

SaschaMann:
> but I don't think there's anything relevant in there unfortunately.

Usually it shows how those expressions are evaluated. The syntax itself is correct (if you do a code search, you can find plenty of workflow files where it's used in that way). Perhaps it needs to be at a step, not a job level, and you have to add it to each `- uses: ...` step? The example in the docs looks similar to what you have, though.

christopher-dG:
So it didn't work even if I just hardcoded `continue-on-error: true`. I'm wondering if it's a GitHub bug because this PR started out with some actual failing jobs which were not `continue-on-error`, so maybe those are what are preventing the status from being green. I'm just gonna merge it and if it still doesn't work I'll come up with something else.

PR #209, opened 2020-08-07
use dynamic path to make sysimg happy

Roger-luo:
currently, if a system image is compiled that contains PkgTemplates, the `Template` functionality won't be redistributable, since it is using a global constant that get evaluated to be a very specific path on the build machine but not the user's actual path during compilation. Then you maybe get something like MIT license not found error.

since TEMPLATES_DIR is only used for `default_files` I think it's OK to just use `pathof`.

### DIFF
--- a/src/plugin.jl
+++ b/src/plugin.jl
@@ -106,7 +106,7 @@ abstract type FilePlugin <: Plugin end
 Return a path relative to the default template file directory
 (`PkgTemplates/templates`).
 """
-default_file(paths::AbstractString...) = contractuser(joinpath(dirname(dirname(pathof(PkgTemplates))), "templates", paths...))
+default_file(paths::AbstractString...) = joinpath(dirname(dirname(pathof(PkgTemplates))), "templates", paths...)
 
 """
     view(::Plugin, ::Template, pkg::AbstractString) -> Dict{String, Any}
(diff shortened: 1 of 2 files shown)


Roger-luo:
Before:

```julia
julia> PkgTemplates.default_file("licenses", "MIT")
"/Users/runner/.julia/packages/PkgTemplates/aXRp5/templates/licenses/MIT"
```

After:

```jl
julia> PkgTemplates.default_file("licenses", "MIT")
"/Users/roger/.julia/packages/PkgTemplates/aXRp5/templates/licenses/MIT"
```

src/plugin.jl
```diff
@@ -105,9 +104,9 @@ abstract type FilePlugin <: Plugin end
     default_file(paths::AbstractString...) -> String
 
 Return a path relative to the default template file directory
-(`$(contractuser(TEMPLATES_DIR))`).
+(`PkgTemplates/templates`).
 """
-default_file(paths::AbstractString...) = joinpath(TEMPLATES_DIR, paths...)
+default_file(paths::AbstractString...) = joinpath(dirname(dirname(pathof(PkgTemplates))), "templates", paths...)
```
christopher-dG, line 109:
```suggestion
function default_file(paths::AbstractString...)
    return joinpath(dirname(dirname(pathof(PkgTemplates))), "templates", paths...)
end
```

christopher-dG:
Great, thank you!

Roger-luo:
Hi @christopher-dG , just wondering if there could be a patch version for this, since one of my package depends on this issue.

christopher-dG:
Sorry! For sure tomorrow, hoping to get #208 in there.

PR #212, opened 2020-08-11
Roll our own continue-on-error

christopher-dG:
Following the adventures of #208 and seeing that it wasn't magically fixed when I merged it, I think it's easiest to just do our own thing.


DilumAluthge:
Personally I think it would be better for us to figure out out how to make continue on error work.

I'll spend a little time on this today.

PR #213, opened 2020-08-11
Allow test failures on Julia nightly

DilumAluthge:
Closes #212

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,6 @@ jobs:
           - windows-latest
         arch:
           - x64
-    continue-on-error: ${{ matrix.version == 'nightly' }}
     steps:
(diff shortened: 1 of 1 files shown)


DilumAluthge:
Looks like this is working correctly.

@christopher-dG 

@SaschaMann

SaschaMann:
I don't like this as a default. It hides the build status of nightly builds completely, so you have to manually check them. I think it's better to have this as an option.

(Personally I don't mind failing checks but set the builds that must pass as required instead)

christopher-dG:
This is only for this repo, not for generated packages. Cassette is a test dependency which often fails on nightly.

christopher-dG:
I wonder why the coverage is so much worse... but oh well.

PR #214, opened 2020-08-11
Disable inlining during the tests

DilumAluthge:
This might give more accurate code coverage information.


DilumAluthge:
Okay... apparently disabling inlining breaks the PkgTemplates test suite.

PR #222, opened 2020-08-19
Fix Manifest.toml entry in gitignore

### LINKED ISSUE #220: Should default `.gitignore` have `/Manifest.toml`  instead of `Manifest.toml`?
The default `.gitignore` will ignore `Manifest.toml` files
```julia
julia> Template(dir=".")("demo.jl")
[ Info: Running prehooks
[ Info: Running hooks
 Activating environment at `~/repos/demo/Project.toml`
   Updating registry at `~/.julia/registries/General`
No Changes to `~/repos/demo/Project.toml`
No Changes to `~/repos/demo/Manifest.toml`
 Activating environment at `~/repos/Project.toml`
[ Info: Running posthooks
[ Info: New package is at /Users/nick/repos/demo

shell> cat ./demo/.gitignore
Manifest.toml
```
But many repos choose to commit (not ignore) `docs/Manifest.toml`, and so would prefer to only ignore `/Manifest.toml`. In fact, PkgTemplates itself does this:
https://github.com/invenia/PkgTemplates.jl/blob/97a062fd4fb798e1172149843d7098e5ec3ba619/.gitignore#L5

I couldn't see an easy way to override this either, e.g. `Git(ignore=["/Manifest.toml"])` results in having `/Manifest.toml` _and_ `Manifest.toml` in the `.gitignore`

Can we change the default to `/Manifest.toml`? 
Alternatively, can we expose an easy way to change this (if there's not one already)?

christopher-dG:
Fixes #220

### DIFF
--- a/src/plugins/git.jl
+++ b/src/plugins/git.jl
@@ -39,12 +39,7 @@ end
 
 # Try to make sure that no files are created after we commit.
 priority(::Git, ::typeof(posthook)) = 5
-
-function gitignore(p::Git)
-    ignore = copy(p.ignore)
-    p.manifest || push!(ignore, "Manifest.toml")
-    return ignore
-end
+gitignore(p::Git) = p.ignore
 
 function validate(p::Git, t::Template)
     if p.gpgsign && !git_is_installed()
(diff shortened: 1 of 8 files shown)


nickrobinson251 (approved):
Thanks!

PR #240, opened 2020-10-29
Update for the new TagBot

DilumAluthge:


DilumAluthge:
@christopher-dG What do you think about merging this PR, and then registering a new version of PkgTemplates?

That way, we can make sure that the new TagBot is working in production on the General registry.

christopher-dG:
Sure, why not.

Issue #233, opened 2020-09-20
Deal with new default branch name


iamed2:
https://github.com/github/renaming

New repos on GitHub will have their default branch set to "main" instead of "master" starting October 1, 2020. 

I think the best thing to do would be to make this change now, as having an alternate default branch name is already well-supported.

mschauer:
The copy paste-instructions for making a new repo on GitHub
```
git remote add origin https://github.com/mschauer/MitosisStochasticDiffEq.jl.git
git branch -M main
git push -u origin main
```
now change the default branch underhandedly, which creates problems with the remote already added. :-/

simeonschaub:
The Documenter template will also need to respect the specified branch, right now Documenter defaults to only deploying docs when on the branch `master`.

christopher-dG:
Let's try to figure out what actually needs doing:

```
$ rg master src templates
src/plugins/ci.jl
136:    "https://travis-ci.com/{{{USER}}}/{{{PKG}}}.jl.svg?branch=master",
310:        "https://{{{HOST}}}/{{{USER}}}/{{{PKG}}}.jl/badges/master/pipeline.svg",
315:        "https://{{{HOST}}}/{{{USER}}}/{{{PKG}}}.jl/badges/master/coverage.svg",
316:        "https://{{{HOST}}}/{{{USER}}}/{{{PKG}}}.jl/commits/master",

src/plugins/license.jl
9:  [here](https://github.com/invenia/PkgTemplates.jl/tree/master/templates/licenses).

src/plugins/coverage.jl
21:    "https://codecov.io/gh/{{{USER}}}/{{{PKG}}}.jl/branch/master/graph/badge.svg",
43:    "https://coveralls.io/repos/github/{{{USER}}}/{{{PKG}}}.jl/badge.svg?branch=master",
44:    "https://coveralls.io/github/{{{USER}}}/{{{PKG}}}.jl?branch=master",

templates/appveyor.yml
22:    - master

templates/cirrus.yml
12:    - sh -c "$(fetch https://raw.githubusercontent.com/ararslan/CirrusCI.jl/master/bin/install.sh -o -)"

templates/gitlab-ci.yml
50:    - master
```

- The occurrences in `license.jl` and `cirrus.yml` don't need to be changed. 
- The badge URLs in the CI and coverage plugins will need to depend on the default branch name, as will the entries in the AppVeyor and GitLab CI configs.
[...]

PR #242, opened 2020-11-04
allow specifying devbranch for Documenter

simeonschaub:
I am not sure, whether this is the right logic for getting the default branch from `Git`, maybe we need to do something smarter here, as pointed out in https://github.com/invenia/PkgTemplates.jl/issues/233#issuecomment-721442945:

> The Git plugin itself already respects Git's `init.defaultBranch` setting, but it doesn't actually store the value anywhere so other plugins can't know what it is. Maybe it can be made to store the default branch as a field, and then other plugins can access the value through it. That does mean we have to choose a "default default" branch, for cases where `init.defaultBranch` isn't set. Git's default is still `master`, but we should probably use `main`.

This should be mergeable without #233, but #233 will require something like this anyways. Also still needs tests.


src/plugins/documenter.jl
```diff
@@ -55,6 +55,7 @@ or `Nothing` to only support local documentation builds.
   for [`TravisCI`](@ref) and [`GitLabCI`](@ref), respectively.
   If set to `nothing`, no canonical URL is set.
 - `makedocs_kwargs::Dict{Symbol}`: Extra keyword arguments to be inserted into `makedocs`.
+- `devbranch::Union{String, Nothing}`: Branch that will trigger the deployment of the site.
```
christopher-dG, line 58:
```suggestion
- `devbranch::Union{AbstractString, Nothing}`: Branch that will trigger docs deployment.
```

src/plugins/documenter.jl
```diff
@@ -112,6 +124,11 @@ badges(::Documenter{GitLabCI}) = Badge(
     "https://{{{USER}}}.gitlab.io/{{{PKG}}}.jl/dev",
 )
 
+function default_branch(t::Template)
```
christopher-dG, line 127: This can actually go somewhere in `template.jl`, I'll be using this in a few places.

src/plugins/documenter.jl
```diff
@@ -121,6 +138,7 @@ view(p::Documenter, t::Template, pkg::AbstractString) = Dict(
     "PKG" => pkg,
     "REPO" => "$(t.host)/$(t.user)/$pkg.jl",
     "USER" => t.user,
+    "DEVBRANCH" => something(p.devbranch, default_branch(t)),
```
christopher-dG, line 141:
```suggestion
    "BRANCH" => p.devbranch === nothing ? default_branch(t) : p.devbranch,
```

I'm just calling this `BRANCH` in all the other views that need it for consistency.
Also, `something` will error if both are `nothing`, but it's actually not a problem.

templates/docs/make.jl
```diff
@@ -30,5 +30,8 @@ makedocs(;
 
 deploydocs(;
     repo="{{{REPO}}}",
+{{#DEVBRANCH}}
+    devbranch="{{{DEVBRANCH}}}",
+{{/DEVBRANCH}}
```
christopher-dG, line 35:
```suggestion
{{#BRANCH}}
    devbranch="{{{BRANCH}}}",
{{/BRANCH}}
```

christopher-dG (commented):
Looks great! Just a few nitpicks.

The failing tests are: 

- interactive.jl: [this line](https://github.com/invenia/PkgTemplates.jl/blob/master/test/interactive.jl#L172) just needs to be changed to do `DOWN^2` instead of just `DOWN`.
- `reference.jl`: Set a custom `devbranch` in the `Documenter` instance of `WackyOptions`, then run the tests in a REPL. It should prompt you to accept some changes to the reference files where they differ, but it's been reported that that feature only works on my machine...

test/reference.jl
```diff
@@ -114,11 +114,12 @@ end
             Codecov(; file=STATIC_TXT),
             CompatHelper(; cron="0 0 */3 * *"),
             Coveralls(; file=STATIC_TXT),
-            Documenter{GitLabCI}(;
+            Documenter{GitHubActions}(;
```
simeonschaub, line 117: I had to change that, because PgkTemplates does not generate the deployment step in `docs/make.jl` for `GitLabCI`.

christopher-dG: This is fine, I will add another testset for `Documenter{GitLabCI}` myself later.

christopher-dG (approved):
Looks great, thanks! Provided that CI passes this is good to go.

PR #243, opened 2020-11-04
Set default branch explicitly

### LINKED ISSUE #233: Deal with new default branch name

christopher-dG:
Closes #233 

The `Git` plugin will set the default branch with the following priority:

- The `branch` keyword to the `Git` constructor
- The Git setting `init.defaultBranch`
- `"main"`

And all the other plugins can use that where they need to refer to the default branch.

This needs the `default_branch` from #242.

This adds a bit of awkwardness for testing with preexisting Git configurations that set `init.defaultBranch` to something other than `main`, since a bunch of tests assume that it'll be unset. I added `with_clean_gitconfig` to sort of solve that, but the mocking is too slow to use on the reference tests so they're just disabled in that case.


nickrobinson251:
bump :)

abelsiqueira:
I would love to see this PR merged soon. Specifically, I'm preparing a tutorial and noticed that `devbranch="main"` is missing from `docs/make.jl`. I'm also keeping an eye at https://github.com/JuliaDocs/Documenter.jl/issues/1443.

christopher-dG:
If anyone has time to take over, they should feel free -- I currently do not, unfortunately.

abelsiqueira:
I could try to understand and fix this PR, but it is massive, and I don't know the project well enough. So I would probably just get frustrated along the way.
Instead, I could try to just fix the `make.jl` part. Is that of interest? At least this would push something over the `main` branch border and might be the kickstart for the rest of the changes.

By the way, my concern here is that I'm using `PkgTemplates` with mostly default options in [my tutorial](https://abelsiqueira.github.io/julia-tutorial-at-nlesc/pages/sequgen/), which creates a `git` repo with `main` branch, but Documenter.jl still uses `master`, so the default is not consistent.

timholy:
Superseded by #316, please close (I forgot to put that in my PR OP).

PR #248, opened 2020-11-18
Remove `L` prefix before line segment of repo URL

nickrobinson251:
i don't know how to explain this one... this `L` has been here a long time, but it seems unnecessary as it always results in URLs like `https://github.com/etc..../foo.jl#LL1-2` with the double `LL`. Here's an example result of this change:
```diff
-https://github.com/invenia/PkgTemplates.jl/blob/89e6dd5e39641c8d40b2492d307e361a83b6dd0f/src/plugin.jl#LL155-L162
+https://github.com/invenia/PkgTemplates.jl/blob/89e6dd5e39641c8d40b2492d307e361a83b6dd0f/src/plugin.jl#L155-L162
```
I tested with Documenter 0.23, 0.24 and 0.25 and it always gives this same `LL` result. 

As you can test for yourself using the two example links above, both work i.e. both link correctly to the source code of the `PkgTemplates.tags(::Plugin)` function.

However, the `LL` doesn't (or doesn't any longer??) work on GitLab. I don't have an open-source GitLab repo where i can show this, but somthing like `https://gitlab.com/etc..../foo.jl#LL1-2` does not highlight lines 1-2 whereas `https://github.com/etc..../foo.jl#L1-2` (with the single `L`) does highlight the intended lines.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "PkgTemplates"
 uuid = "14b8a8f1-9102-5b29-a752-f990bacb7fe1"
 authors = ["Chris de Graaf", "Invenia Technical Computing Corporation"]
-version = "0.7.13"
+version = "0.7.14"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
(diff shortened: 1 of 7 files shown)


christopher-dG:
Turns out this has been outdated since Documenter 0.15! https://juliadocs.github.io/Documenter.jl/v0.14/lib/public/#Documenter.makedocs

christopher-dG (approved):
Reference tests need updating. Thanks!

nickrobinson251:
i don't have merge rights, but hopefully this is good to go :)

PR #249, opened 2020-11-24
Adding a `prefix` keyword argument to Git plugin

manuelbb-upb:
My university offers a custom git(lab) server.
When using ssh access, this server requires the remote URL to start with "irb-git@"  instead of the usual "git@".
As there was no way to change the default behavior of the Git plugin, I added a keyword argument `prefix`.
If is left unset the default prefix is used (either `"https://"` or `"git@"`).
I can now use `Git(; ssh = true, prefix="irb-git@", … )` and all works fine.

An alternative approach to resolve similar problems could be to let the user define the complete remote url.

### DIFF
--- a/src/plugins/git.jl
+++ b/src/plugins/git.jl
@@ -5,6 +5,7 @@
         email=nothing,
         branch=nothing,
         ssh=false,
+        prefix=nothing,
         jl=true,
         manifest=false,
         gpgsign=false,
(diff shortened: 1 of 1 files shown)


christopher-dG:
I'm going to recommend that you write your own plugin for this, since this is kind of Git setup is definitely rare and I'm not sure that PkgTemplates needs to accommodate it by default. Thankfully, this is basically the reason that the plugin system exists!

Here's something that works:

```jl
using LibGit2: LibGit2, GitRemote, GitRepo, set_remote_url
using PkgTemplates: PkgTemplates, Git, Plugin, Template, @plugin, @with_kw_noshow

@plugin struct RemotePrefix <: Plugin
    prefix::String = ""
end

function PkgTemplates.posthook(p::RemotePrefix, ::Template, pkg_dir::AbstractString)
    LibGit2.with(GitRepo(pkg_dir)) do repo
        set_remote_url(repo, "origin", p.prefix * remote_url(repo)
    end
end

remote_url(repo) = LibGit2.with(LibGit2.url, LibGit2.get(GitRemote, repo, "origin"))

t = Template(; plugins=[Git(; ssh=true), RemotePrefix("irb-")])
t("MyPkg")
```

PR #250, opened 2020-11-30
allow changing default template dir

Roger-luo:
this follows #209  the current implementation is still not relocatable - the directory PkgTemplate may not exist if it is compiled into the system image, so one way to get rid of it is to allow reset some of the default paths to new places.

### DIFF
--- a/src/plugin.jl
+++ b/src/plugin.jl
@@ -1,4 +1,5 @@
 const DEFAULT_PRIORITY = 1000
+const DEFAULT_TEMPLATE_DIR = Ref{String}(joinpath(pkgdir(PkgTemplates), "templates"))
 
 """
     @plugin struct ... end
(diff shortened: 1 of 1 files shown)


src/plugin.jl
```diff
@@ -107,7 +108,7 @@ Return a path relative to the default template file directory
 (`PkgTemplates/templates`).
 """
 function default_file(paths::AbstractString...)
-    return joinpath(dirname(dirname(pathof(PkgTemplates))), "templates", paths...)
+    return joinpath(DEFAULT_TEMPLATE_DIR[], paths...)
```
christopher-dG, line 111: This can be made into a one line function now

src/plugin.jl
```diff
@@ -1,4 +1,5 @@
 const DEFAULT_PRIORITY = 1000
+const DEFAULT_TEMPLATE_DIR = Ref{String}(joinpath(pkgdir(PkgTemplates), "templates"))
```
christopher-dG, line 2:
```suggestion
const DEFAULT_TEMPLATE_DIR = Ref(joinpath(pkgdir(PkgTemplates), "templates"))
```

christopher-dG (commented):
Thanks! Just a couple of little style nitpicks. 

Edit: also needs to go back to `pathof` to be 1.0 compatible.

Roger-luo:
@christopher-dG bump

christopher-dG:
Sorry, and thanks! I am hoping to get a new release out with some other PRs on Monday.

PR #251, opened 2020-12-01
GitHub Actions: Put the doctests into a separate job

DilumAluthge:
This allows e.g.:
1. Docs to build even if there is a doctest failure. Doctests can fail between minor versions of Julia, so doctests failures should not necessarily preclude the docs from building.
2. Using different versions of Julia for the docs vs. the doctests (we just did this in LoopVectorization.jl)


DilumAluthge:
Closing this in favor of a larger pull request that I will open soon.

PR #252, opened 2020-12-02
Reduce number of CI jobs

rikhuijzer:
At https://github.com/JuliaLang/Example.jl/pull/55, Fredrik Ekre mentions that PkgTemplates advises quite a lot of jobs by default. I agree and would also argue that most repositories only run a few Linux jobs. In this PR, I propose to comment out some of the jobs, so that users can decide for themselves. This is just a suggestion; let me know if I should update the Travis and Appveyor and tests accordingly. 

The change of filename is just because `ci.yml` looks a bit weird next to `CompatHelper.yml` and `TagBot.yml`, see, for example, https://github.com/rikhuijzer/Gadfly.jl/tree/master/.github/workflows.


christopher-dG:
This PR doesn't do what you think it does, you're just changing PkgTemplates' CI configs, not the ones it generates.

To do the thing you want to do, you just need to change the default value of `osx` and `windows` to `false` in the `GitHubActions` struct definition.

Personally, I'm on board with this, I generally only test Linux myself unless my package does a lot of FS stuff or other platform-dependent stuff. @nickrobinson251 could I get your take on behalf of Invenia?

rikhuijzer:
Okay, thanks for the pointers. :smile: I'll now wait for the reply from Invenia('s representative)

nickrobinson251:
Yeah, i think templates defaulting to running on Linux only is fine! 

i.e.
```julia
        linux=true,
        osx=false,
        windows=false,
```
Probably want `GitHubActions` and `TravisCI` to be consistent on this?

src/plugins/ci.jl
```diff
@@ -14,11 +14,11 @@ const EXTRA_VERSIONS_DOC = "- `extra_versions::Vector`: Extra Julia versions to
 
 """
     GitHubActions(;
-        file="$(contractuser(default_file("github", "workflows", "ci.yml")))",
+        file="$(contractuser(default_file("github", "workflows", "CI.yml")))",
```
christopher-dG, line 17: If you're going to change this (I have no objections), you need to also change it for `destination` and also in the actual default values at lines 48-49.

rikhuijzer: Yes, thank you again. I could have easily spotted that (again), and was a bit careless. I'll try to do better on the next commit.

christopher-dG:
Ok, looking good. The last step is to update the tests. You can move the existing `ci.yml` reference files to their new `CI.yml` paths:

```
test/fixtures/AllPlugins/.github/workflows/ci.yml
test/fixtures/WackyOptions/.github/workflows/ci.yml
test/fixtures/DocumenterGitHubActions/.github/workflows/ci.yml
```
And then run the unit tests from the Pkg REPL mode, you should be prompted to accept the changes to those files (although some people have reported that this bit doesn't quite work).

rikhuijzer:
Clear explanation. Thank you. 

I've ran the tests and got prompted to change two files (pressed `y` twice). Then, I ran the tests again and everything passed.

src/plugins/ci.jl
```diff
@@ -93,8 +93,8 @@ end
     TravisCI(;
         file="$(contractuser(default_file("travis.yml")))",
         linux=true,
-        osx=true,
-        windows=true,
+        osx=false,
+        windows=false,
```
nickrobinson251, line 97: Needs changing in the code below too I think

rikhuijzer: I'm afraid I don't understand this suggestion. I take it that in some more places `osx=true` should be `osx=false`? I did a grep search on `osx=true` in the branch and found no matches.

nickrobinson251: sorry, should have been more precise. this only changes the docstring, we need to also change the code for `TravisCI` below at Line 121 - 123
https://github.com/invenia/PkgTemplates.jl/pull/252/files#diff-d7163122af7f8ded3f051c46b47f1e9d78841134eff0368b308067fb607712eaR121-R123

Perhaps grepping for just `osx` will catch other cases, if there are any? :)

rikhuijzer: Well spotted. I've changed it and am running the tests now locally.

christopher-dG (commented):
Something odd appears to have happened with `WackyOptions`' GHA config, there is a file both at `.github/CI.yml` and at `.github/workflows/CI.yml` (the latter is the correct path).

rikhuijzer:
> Something odd appears to have happened with `WackyOptions`' GHA config, there is a file both at `.github/CI.yml` and at `.github/workflows/CI.yml` (the latter is the correct path).

Probably me making a mistake when using `mv`.

PR #253, opened 2020-12-04
MassInstallAction: Install the GitHubActions workflow on this repository

fchorney:
This pull request sets up the GitHubActions workflow on this repository. cc: @fchorney

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -61,9 +61,8 @@ mktempdir() do dir
 
                 # Quite a bit of output depends on the Julia version,
                 # and the test fixtures are made with Julia 1.5.
-                # These tests fail on 32-bit
                 # TODO: Keep this on the latest stable Julia version.
-                if VERSION.major == 1 && VERSION.minor == 5 && Sys.WORD_SIZE == 64
+                if VERSION.major == 1 && VERSION.minor == 5
                     include("reference.jl")
                 else
                     @info "Skipping reference tests" VERSION
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -61,8 +61,9 @@ mktempdir() do dir
 
                 # Quite a bit of output depends on the Julia version,
                 # and the test fixtures are made with Julia 1.5.
+                # These tests fail on 32-bit
                 # TODO: Keep this on the latest stable Julia version.
-                if VERSION.major == 1 && VERSION.minor == 5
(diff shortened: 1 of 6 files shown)


christopher-dG:
@fchorney this one can be closed, right?

fchorney:
@christopher-dG do we not want to update the GHA CI scripts?

christopher-dG:
I was mostly thinking that this repo didn't need the Slack and nightly stuff since I'm mostly the one who does maintenance, but if you guys want it that's no problem (especially if it helps you guys maintain it too!).

fchorney:
> I was mostly thinking that this repo didn't need the Slack and nightly stuff since I'm mostly the one who does maintenance, but if you guys want it that's no problem (especially if it helps you guys maintain it too!).

ah good point. I'll see what people think.

test/runtests.jl
```diff
@@ -61,8 +61,9 @@ mktempdir() do dir
 
                 # Quite a bit of output depends on the Julia version,
                 # and the test fixtures are made with Julia 1.5.
+                # These tests fail on 32-bit
```
mattBrzezinski, line 64: What's the reason for these failing on 32-bit? Do we have an issue logged for it? Or is it even worth logging / already know elsewhere?

christopher-dG: I wasn't aware of these failures, and I can't really think of why they would break there. So it's a good test to have!

fchorney: I am currently unsure why they fail on 32-bit. They seem to be failing in a spot that's meant to fail but works fine on 64-bit? I didn't look into it super deep

fchorney: for reference (hah) it was failing here only on 32-bit for whatever reason https://github.com/invenia/PkgTemplates.jl/blob/master/test/reference.jl#L50

christopher-dG: If you can find a link to the job logs, that would help a lot (or just enable those tests again for 32-bit)

fchorney: Should be in this run https://github.com/invenia/PkgTemplates.jl/pull/253/checks?check_run_id=1526431542

fchorney: Thanks for the fix!

mjram0s (approved):

mattBrzezinski:
:+1:

PR #254, opened 2020-12-04
switch to Configurations to support better @plugin macro and template serailization in TOML

Roger-luo:
This PR removes `Parameters` and use `Configurations` instead, since I think the goal of [Configurations](https://github.com/Roger-luo/Configurations.jl) fits better for the PkgTemplate use case, and there is not much user interface change, all the old interfaces are the same including `@plugin` macro. 

But you don't need to import `@with_kw_noshow` anymore, only use `@plugin` is sufficient since you can select your own code generators in Configurations. I think It makes the implementation of `Plugin` simpler and better and reduces the maintenance pressure of this package since I'm also going to help Pluto and Comonicon to switch to this configuration package - most of the code in Pluto, Comonicon's configuration structs and PkgTemplates' Plugin are actually similar. So I think this brings a much simpler implementation of `Plugin`.

A glance at some new features automatically (or by slightly overloading some interfaces of `Configurations`) enabled by using Configurations

#### Color, Compact, Copy Pastable Printing

I think currently it is a bit painful to read when all the fields are printed no matter it's just default value or not, now default values are automatically hidden and colors are used to help distinguish between kwargs, Plugin types and values. And you cannot copy pasta the printings back to REPL - this is also not convenient for sharing.

Before

image

After

image

#### Serialization to TOML files and potentially many other formats

now you can use `Configurations.to_toml` function to convert your templates to a TOML file easily. This makes sharing templates and modifying templates a lot easier and flexible. Also make things easier to read.

image

can you can read a template from a TOML file by `Configurations.from_toml`

image

since you can also convert a `Template` to `OrderedDict` using `to_dict`, it can be serialized to other format you like as well!


Roger-luo:
I'll add some more tests if people like this PR.

christopher-dG:
Holy cow, this is awesome. It'll take me a long time to review this but I definitely support the idea.

Roger-luo:
cool, I'll fix a few issues left first then:

- [ ] do not expand homedir in serialized format or printings
- [ ] fix 1.0 compatibility
- [ ] add some tests for the serialization part

I think I can I assume the rest part should be covered by current test, but I could be wrong.

Roger-luo:
I realize we need to have https://github.com/JuliaLang/TOML.jl/issues/13 fixed to be able to disable and use the default field of specified plugins. So don't merge this until that issue is fixed.

Roger-luo:
TBH I have no idea why the `mock` test errors in 1.0, if you remove the mock, it seems to work fine, but with the `mock`, I got the following locally. This is fine on other julia versions. I think @christopher-dG  could review this now. I've fixed the TOML.jl issue above myself too (so I mark the test here as broken for now, and we could just change it back to `@test` when my PR is merged to TOML)

```julia
julia> mock(PT.version_of => _p -> v"1.2.3") do _i
                   with_pkg(t) do pkg
                       pkg_dir = abspath(expanduser(joinpath(t.dir, pkg)))
                       LibGit2.with(GitRepo(pkg_dir)) do repo
                           commit = GitCommit(repo, "HEAD")
                           @test occursin("PkgTemplates version: 1.2.3", LibGit2.message(commit))
                       end
                   end
               end
[ Info: Running prehooks
[ Info: Running hooks
```

christopher-dG:
Setting myself a reminder to review this weekend.

christopher-dG:
Sorry I've let this slip again. @nickrobinson251 any chance someone from Invenia would like to review? A big change in internals like this might be a good time for maintainership to pass over more formally to current Invenians.

nickrobinson251:
> @nickrobinson251 any chance someone from Invenia would like to review? A big change in internals like this might be a good time for maintainership to pass over more formally to current Invenians.

Sure! I'll bring it up this week to see who'd be interested :) 

fyi @nick-thiessen and @omus

nick-thiessen:
At the moment no one is free to review this, but in the next couple weeks someone might become available and possibly be available to take over maintainership. I'll have to have some conversations about who is interested.

gdalle:
Currently trying to clean up the repo a little. Is this PR still relevant @Roger-luo?

Roger-luo:
Not anymore I've implemented this in rust

gdalle:
My question was badly phrased, I was rather wondering whether it could be brought up to speed with the latest changes to the package?

Roger-luo:
I mean I don't even use this package anymore... I have no motivation to keep working on this PR, at least for myself, my own CLI tool is much easier to use than this package. https://discourse.julialang.org/t/ann-the-ion-command-line-for-julia-developers-written-in-rust/94495

gdalle:
Okay thanks. Still, I might revisit your ideas and try to switch to a better serialization at some point!

PR #257, opened 2020-12-12
Two modifications to the templates for GitHub Actions CI

DilumAluthge:
This pull request makes two modifications to the templates for GitHub Actions CI. These modifications are motivated by feedback that I have observed from various users.

### 1. Separate the single `docs` job into two separate jobs for (1) running the doctests and (2) building and deploying the docs

There are many reasons why doctests might be failing. For example, between Julia 1.5 and 1.6, there have been several changes to printing (e.g. something that is printed like `Array{Int64,1}` in Julia 1.5 is printed like `Vector{Int64} (alias for Array{Int64, 1})` in Julia 1.6). Therefore, it is very reasonable for users to want their docs to build and deploy even if their doctests are failing.

Currently, we do everything in a single `docs` job that runs the doctests prior to building and deploying the docs. Therefore, if the doctests fail, the docs are not built.

With this PR, the docs are built and deployed in the `docs` job, and the doctests are run in the `doctests` job. The docs will be built and deployed even if the doctests fail.

### 2. Move the tests on Julia nightly into a separate workflow file, which allows the user to generate separate README badges for `CI` versus `CI (Julia nightly)`

Many users have asked if GitHub Actions has a option analogous to this feature in Travis CI:
```yaml
allow_failures:
  - julia: nightly
```

Unfortunately, GitHub Actions does not have an analogous feature. There are several alternatives, but none of them are good:
1. Use `continue-on-error`. Unfortunately, if you do this, GitHub Actions will make it look like your CI is always passing on Julia nightly. But this is not what people want. People want to know whether or not their CI passes on Julia nightly, but they don't want failures on Julia nightly to make their overall README CI badge status be "failing".
2. Don't test on Julia nightly. This is not an option for many people.
3. The current approach. Unfortunately, with the current approach, if your CI is failing on Julia nightly, but passing on all other Julia versions, the CI badge in your README will still say "failing".

This pull request moves the Julia nightly job to a separate CI file. Therefore, the `CI` badge will only apply to non-nightly jobs. There will be a separate `CI (Julia nightly)` badge for the nightly jobs.

### DIFF
--- a/src/plugins/ci.jl
+++ b/src/plugins/ci.jl
@@ -22,7 +22,7 @@ const EXTRA_VERSIONS_DOC = "- `extra_versions::Vector`: Extra Julia versions to 
         x64=true,
         x86=false,
         coverage=true,
-        extra_versions=$DEFAULT_CI_VERSIONS,
+        extra_versions=$DEFAULT_CI_VERSIONS_NO_NIGHTLY,
     )
 
 Integrates your packages with [GitHub Actions](https://github.com/features/actions).
@@ -53,7 +53,7 @@ $EXTRA_VERSIONS_DOC
     x64::Bool = true
     x86::Bool = false
     coverage::Bool = true
-    extra_versions::Vector = DEFAULT_CI_VERSIONS
+    extra_versions::Vector = DEFAULT_CI_VERSIONS_NO_NIGHTLY
 end
 
 source(p::GitHubActions) = p.file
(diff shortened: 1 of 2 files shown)


DilumAluthge:
@christopher-dG Is there a way to tell the `GitHubActions` plugin to generate two files (`ci.yml` and `ci-julia-nightly.yml`) instead of just one file (`ci.yml`)?

DilumAluthge:
I need to think about this more.

PR #262, opened 2021-01-10
Return the package directory

kskyten:
Returning the package directory makes it easy to change to the newly created package with

```julia
cd(MyTemplate("NewPkg"))
```


gdalle:
I think this is a good idea. Could you maybe document the return behavior somewhere?

gdalle (approved):

PR #263, opened 2021-01-18
Fix CI Tests and Artifact Cache Issues

fchorney:
Use CLI script to automatically replace the 1.3 test with a 1.5 test (if necessary), as well as use new cache steps to cache the julia build artifacts


mattBrzezinski:
Looks like we need to add some ENV VARs for fake Git username or something for the CI to pass here?

fchorney:
> Looks like we need to add some ENV VARs for fake Git username or something for the CI to pass here?

oh right, this might be one of the only situations where those were actually needed. I'll put them back

PR #264, opened 2021-01-23
some changes to allow for more customization of badges

ExpandingMan:
This PR includes a couple of small changes to make plugins more easily customizable, particularly the printing of badges in the README.

First, I have dropped the restriction to `DeployStyle` from the type parameter of `Documenter`.  The reason for this is that, as `DeployStyle` is a type union, a user cannot create a new plugin type and provide it as a type parameter to `Documenter`.  This is needed to, e.g. change the badges returned by the `Documenter` plugin.

Next, I have added the `badge_order::Vector{DataType}` field to the `Readme` struct.  This allows users to set their own badge order.  The original default order remains the default, but I have changed the name of the function that creates this order to `default_badge_order` for clarity.

I don't think there should be any test errors here on 1.5, but I worked on this on 1.6 in which I get some rather inscrutable Cassette (via SimpleMock) errors when testing.  Hopefully there will be no errors in the automated unit tests, but if there are, I'll probably have to look into the Cassette errors before this goes anywhere.

Thanks!

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "PkgTemplates"
 uuid = "14b8a8f1-9102-5b29-a752-f990bacb7fe1"
 authors = ["Chris de Graaf", "Invenia Technical Computing Corporation"]
-version = "0.7.14"
+version = "0.7.15"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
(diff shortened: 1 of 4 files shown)


nicoleepp (approved):

morris25 (approved):

ExpandingMan:
Great, thanks!

Issue #247, opened 2020-11-17
Support turning off badges for plugins


nickrobinson251:
e.g. to use `GitLabCI` but not have a badge in the readme

Issue #246, opened 2020-11-17
Support custom domain for docs badge


nickrobinson251:
https://github.com/invenia/PkgTemplates.jl/blob/89e6dd5e39641c8d40b2492d307e361a83b6dd0f/src/plugins/documenter.jl#L123-L124

nickrobinson251:
Actually this effects most plugins with badges, since most Ci-related things (coverage, docs etc) can be at custom locations. Also #247 would basically be a workaround for this.

PR #265, opened 2021-02-03
Change GitHubActions badge to match action filename

nickrobinson251:
This changes the badge to depend on the action name, to allow multiple CI actions e.g.
```julia
GihubActions()  # CI 
GithubActions(file=default("JuliaNightly.yml"), destination="JuliaNightly.yml)  # CI on 'nightly' version of Julia 
```
and have a badge for each, like https://github.com/invenia/nameddims.jl/workflows/CI/badge.svg and https://github.com/invenia/nameddims.jl/workflows/JuliaNightly/badge.svg

Questions:
- Do we also need to change the link to `https://github.com/{{{USER}}}/{{{PKG}}}.jl/actions?query=workflow:name`? e.g. so the `JuliaNightly` badge would go to e.g. https://github.com/Invenia/NamedDims.jl/actions?query=workflow:JuliaNightly
- Where/how is best to add a test for this?

(Kinda related to #246 / #247 )

### DIFF
--- a/src/plugins/ci.jl
+++ b/src/plugins/ci.jl
@@ -60,9 +60,9 @@ source(p::GitHubActions) = p.file
 destination(p::GitHubActions) = joinpath(".github", "workflows", p.destination)
 tags(::GitHubActions) = "<<", ">>"
 
-badges(::GitHubActions) = Badge(
+badges(p::GitHubActions) = Badge(
     "Build Status",
-    "https://github.com/{{{USER}}}/{{{PKG}}}.jl/workflows/CI/badge.svg",
+    "https://github.com/{{{USER}}}/{{{PKG}}}.jl/workflows/$(first(splitext(p.destination)))/badge.svg",
     "https://github.com/{{{USER}}}/{{{PKG}}}.jl/actions",
 )
 


src/plugins/ci.jl
```diff
@@ -60,9 +60,9 @@ source(p::GitHubActions) = p.file
 destination(p::GitHubActions) = joinpath(".github", "workflows", p.destination)
 tags(::GitHubActions) = "<<", ">>"
 
-badges(::GitHubActions) = Badge(
+badges(p::GitHubActions) = Badge(
     "Build Status",
-    "https://github.com/{{{USER}}}/{{{PKG}}}.jl/workflows/CI/badge.svg",
+    "https://github.com/{{{USER}}}/{{{PKG}}}.jl/workflows/$(first(splitext(p.destination)))/badge.svg",
     "https://github.com/{{{USER}}}/{{{PKG}}}.jl/actions",
```
nickrobinson251, line 66: i.e. should this change to be like
```suggestion
    "https://github.com/{{{USER}}}/{{{PKG}}}.jl/actions?query=workflow:$(first(splitext(p.destination)))",
```

nickrobinson251:
this would be nice to have regardless, but i see that we can't actually have multiple `GithubAction` CI plugins. Why is that? (i'll open an issue)

The results is thatthe first is chosen, e.g.
```
Template(plugins=[GitHubActions(destination="JuliaNightly.yml"), GitHubActions()])
```
will results in only the JuliaNightly.yml file (and badge etc) being created, and no CI.yml

christopher-dG:
Best way to test this would probably be to change the `destination` here:

https://github.com/invenia/PkgTemplates.jl/blob/238bfc11c037f341eff4889e0424a66d1edf7404/test/reference.jl#L126

christopher-dG:
I think you'll need to change the name of the workflow to match as well:

https://github.com/invenia/PkgTemplates.jl/blob/238bfc11c037f341eff4889e0424a66d1edf7404/templates/github/workflows/CI.yml#L1

nickrobinson251:
> I think you'll need to change the name of the workflow to match

Okay, done. And just switched to reading this from the file itself :)

nickrobinson251:
bump :)

src/plugins/ci.jl
```diff
@@ -60,11 +60,15 @@ source(p::GitHubActions) = p.file
...
 tags(::GitHubActions) = "<<", ">>"
 
-badges(::GitHubActions) = Badge(
-    "Build Status",
-    "https://github.com/{{{USER}}}/{{{PKG}}}.jl/workflows/CI/badge.svg",
-    "https://github.com/{{{USER}}}/{{{PKG}}}.jl/actions",
-)
+function badges(p::GitHubActions)
+    file = read(p.file, String)
+    action_name = strip(first(match(r"name:(.*)", file).captures))
```
christopher-dG, line 65: It's not required that a workflow file has a `name` key, so this is kind of dangerous. Perhaps it can try this but also have a backup?

christopher-dG (changes requested):
Needs to handle workflow files that do not have a `name` key, but otherwise LGTM :)

src/plugins/ci.jl
nickrobinson251: ah, yes GitHub says:

> `name`
>
> The name of your workflow. GitHub displays the names of your workflows on your repository's actions page. **If you omit `name`, GitHub sets it to the workflow file path relative to the root of the repository.**

PR #272, opened 2021-02-08
Set a default time-out of 60 min for GitHub CI

thchr:
The default time-out for Github Actions seems to be 6 hours which is quite a lot (e.g., I ran into this yesterday with two jobs that stalled overnight, eating up 12 hours cumulatively of my "private" time; ouch.)

I doubt a lot projects need CI times longer than, say, 1 hour; and if they do, they're probably aware and happy to edit the default settings. On the other hand, it would be nice to "cap" the damage a stalled job could do to something less than 6 hours. Here, I tentatively just set it to 60 minutes.

In principle, this could be an adjustable option - but I think this is a reasonable default that most people wouldn't need to change until quite far into development.


thchr:
... it's kind of awful that I accidentally started four separate Github Actions on this particular pull request; apologies (was using the GitHub web interface to commit without thinking of the fact that each small commit would set off another CI run).

DilumAluthge:
Bump. Would be great to get this merged.

@christopher-dG

gdalle (approved):

PR #296, opened 2021-06-09
don't use PkgServers in Documenter job

simeonschaub:
The version of the registry served by PkgServers can sometimes lag a
little bit behind, leading to issues such as in JuliaDiff/FiniteDifferences.jl#173.
The `julia-buildpkg` action actually already does this here:
https://github.com/julia-actions/julia-buildpkg/blob/1b0a69ab74709d3707f86249fce805049212961a/action.yml#L22


simeonschaub:
Bump. @DilumAluthge does that look good to you?

oxinabox:
resolved by 
https://github.com/invenia/PkgTemplates.jl/pull/304

PR #297, opened 2021-06-17
Use Mocking instead of SimpleMock

### LINKED ISSUE #287: TypeError causing segfault in Julia 1.6 tests
https://github.com/invenia/PkgTemplates.jl/runs/2198570926?check_suite_focus=true#step:7:187

```
Internal error: encountered unexpected error in runtime:
TypeError(func=:TypeVar, context="upper bound", expected=Type, got=Vararg{Any, N} where N)
jl_type_error_rt at /buildworker/worker/package_linux64/build/src/rtutils.c:119
jl_new_typevar at /buildworker/worker/package_linux64/build/src/builtins.c:1116
TypeVar at ./boot.jl:382
apply_type_tfunc at ./compiler/tfuncs.jl:1231
_jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2237 [inlined]
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2419
jl_apply at /buildworker/worker/package_linux64/build/src/julia.h:1703 [inlined]
do_apply at /buildworker/worker/package_linux64/build/src/builtins.c:670
builtin_tfunction at ./compiler/tfuncs.jl:1488
abstract_call_builtin at ./compiler/abstractinterpretation.jl:826
jfptr_abstract_call_builtin_9925.clone_1 at /opt/hostedtoolcache/julia/1.6.0/x64/lib/julia/sys.so (unknown line)
...
```

mjram0s:
[SimpleMock/Cassette is broken on 1.6](https://github.com/invenia/PkgTemplates.jl/issues/287#issuecomment-808463380) causing the tests to segfault. 

Switching over to use Mocking so package doesn't break on 1.6.

Closes #287


test/git.jl
```diff
@@ -69,7 +69,9 @@
     @testset "Adds version to commit message" begin
         # We're careful to avoid a Pkg.update as it triggers Cassette#130.
         t = tpl(; plugins=[Git(), !Tests])
-        mock(PT.version_of => _p -> v"1.2.3") do _i
+
+        patch = @patch PkgTemplates.version_of(t) = v"1.2.3"
```
mattBrzezinski, line 73: Personally I like putting all patches into a `test/patches.jl` file, naming them their respective function calls.

Maybe a bit of an alternative solution here but this works too. LGTM!

mattBrzezinski (approved):

test/template.jl
```diff
@@ -76,7 +80,9 @@
         foreach((GitHubActions, TravisCI, GitLabCI)) do T
             @test_throws ArgumentError tpl(; plugins=[Documenter{T}()])
         end
-        mock(LibGit2.getconfig => (_k, _d) -> "") do _gc
+
+        patch = @patch LibGit2.getconfig(r, n) = ""
+        apply(patch) do
```
oxinabox, line 85: Is there a reason to define the patch variable that is only used once, rather than `apply(@patch ...) do`?

oxinabox (approved):

test/git.jl
oxinabox: I think it depends how they are used.
In these cases they are only used once so having them inplace lakes more sense.

test/template.jl
mjram0s: I just do it this way because some patches get long and it gets hard to read inside `apply(...)`

test/git.jl
mjram0s: Oh interesting. I will stick to this for now but that is also a good way of doing it.

mattBrzezinski: Yeah these are fine, I think if we start re-using some or expanding out might be something to consider.

PR #303, opened 2021-08-16
MassInstallAction: Install the BlueStyleFormatter workflow on this repository

mattBrzezinski:
This pull request sets up the BlueStyleFormatter workflow on this repository. cc: @mattBrzezinski

### DIFF
--- a/.github/workflows/blue_style_formatter.yml
+++ b/.github/workflows/blue_style_formatter.yml
@@ -1,0 +1,20 @@
+name: Format suggestions
+
+on:
+  pull_request:
+
+jobs:
+  format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@latest
+        with:
+          version: 1
+      - run: |
+          julia  -e 'using Pkg; Pkg.add("JuliaFormatter")'
+          julia  -e 'using JuliaFormatter; format(".", BlueStyle(); verbose=true)'
+      - uses: reviewdog/action-suggester@v1
+        with:
+          tool_name: JuliaFormatter
+          fail_on_error: true
--- a/.github/workflows/blue_style_formatter.yml
+++ b/.github/workflows/blue_style_formatter.yml
@@ -13,7 +13,8 @@ jobs:
           version: 1
       - run: |
           julia  -e 'using Pkg; Pkg.add("JuliaFormatter")'
-          julia  -e 'using JuliaFormatter; format(".", BlueStyle(); verbose=true)'
+          julia  -e 'using JuliaFormatter; format("src/", BlueStyle(); verbose=true)'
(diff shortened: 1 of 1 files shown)


mattBrzezinski:
Modified to only format the `src/` and `test/` directories. Otherwise the check will fail under `templates/` as they are not proper Julia files.

gdalle:
Thanks for the suggestion! I usually prefer testing formatting in `runtests.jl`, cause the review workflow makes lots of small individual suggestions that end up being pretty noisy. Thoughts?

oxinabox:
the advantage of this over doing it in tests is this is all hooked up to review-dog.
Which means it automatically comments on the PR.
And if the PR is from a branch (rather than a fork) it posts suggestions in those comments.
(idk why it can't from a fork)

gdalle:
Either is fine by me, I'm not strongly opinionated.

Another question is which directories do we run the action on?

- `src` is fine
- `test` contains the fixtures, and we may not want to format those

oxinabox:
I actually don't care for autoformatters at all, and think they are more trouble than they are worth, so i have no strong opinions

ViralBShah:
I share the same opinion as @oxinabox but please reopen if we need to do this.

mattBrzezinski:
> I share the same opinion as @oxinabox but please reopen if we need to do this.

We do not, I believe this was one of the packages owned by Invenia at one point and has not been transferred to JuliaCI where folks can do with it as they please.

https://github.com/JuliaRegistries/General/pull/75915/files

ViralBShah:
Fair enough. I'll leave it open.

mattBrzezinski:
> Fair enough. I'll leave it open. I only closed it since there seemed to not have been a decision for a while. Happy for folks who maintain it to decide what should be done here.

To be honest, I think closing it is fine. It's been around for ~2.5years and this was automatically opened via a script. I also would not expect any previous-Invenia employees to be maintaining any Julia packages moving forward as most have moved away from the ecosystem.

gdalle:
speaking of, would it make sense to transfer the BlueStyle repo somewhere else?

mattBrzezinski:
> speaking of, would it make sense to transfer the BlueStyle repo somewhere else?

Could do that, not sure if there's a Julia repo for style guides? There are some Invenia folks floating about that can do _some_ maintenance over the repos but it's a much smaller population now.

oxinabox:
Yes, myself and several others can still maintain it.
We could transfer it, but it is a style guide, it doesn't need much maintenance.
I guess JuliaDiff might make sense as a location for it, since all the ChainRules packages follow it.

gdalle:
>  I guess JuliaDiff might make sense as a location for it, since all the ChainRules packages follow it.

I think anywhere makes more sense than invenia, because leaving it there might give people the impression that it is no longer maintained / useful / used

gdalle:
https://github.com/invenia/BlueStyle/issues/117

PR #304, opened 2021-08-18
Stop using package server as default for docs ci

thomasgudjonwright:


oxinabox (approved):

mjram0s (approved):

iamed2:
Please do not merge branches that do not account for failing tests. You have introduced a state where one CI check will always fail, since you did not update the comparison file at https://github.com/invenia/PkgTemplates.jl/blob/master/test/fixtures/DocumenterGitHubActions/.github/workflows/CI.yml

iamed2:
Why was this done for only docs CI? What makes docs CI different from test CI?

oxinabox:
Tests CI already has it disabled, because it uses the julia-buildpkg action.
https://github.com/julia-actions/julia-buildpkg/blob/master/action.yml#L32
That was added in https://github.com/julia-actions/julia-buildpkg/pull/6

A case could be made that we should use that action here.

PR #315, opened 2021-10-02
Avoid duplication for direct-push access

timholy:
This avoids running CI twice for users who have privileges allowing them
to push directly to the package repo.

The tests fail for me locally at https://github.com/invenia/PkgTemplates.jl/blob/12fb46347be56f2fca954edf24c895f890544c8a/test/interactive.jl#L243
(it fills in my username regardless) so I don't know if this will pass.


timholy:
OK, this seems ready from my perspective. (I should have initially used a draft, sorry about that.)

templates/github/workflows/CI.yml
```diff
@@ -1,7 +1,13 @@
...
-  - push
-  - pull_request
+  create:
+    tags:
+  push:
+    <<#BRANCH>>
+    branches:
+      - <<BRANCH>>
+    <</BRANCH>>
+  pull_request:
```
nickrobinson251, line 10: just to check i'm understanding:
- if a default branch isn't specified (like `Git(; branch = "default")`) then this is the same as before
- but if it is specified, then CI only runs on pushes to that branch, whereas before it would run on any push to the repo, which meant that users with push access had CI run whenever they pushed a branch they were working on... and this led to two CI runs when such a user pushed to the branch and opened a PR?

templates/github/workflows/CI.yml
```diff
@@ -1,7 +1,13 @@
 name: CI
 on:
-  - push
-  - pull_request
+  create:
+    tags:
```
nickrobinson251, line 4: Does this change means CI runs whenever new tags are created? (whereas before it did not run on new tags)

I think that would be a good addition! But i've never seen this `create:` syntax before (and don't see it in [the GHA docs](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions)), so just want to double check this works as expected? 

I wonder if we want `tags: *` instead? (Seems this is what's used e.g. [in Revise.jl CI](https://github.com/timholy/Revise.jl/blob/c82c8a3a04e3cc4e4dc0426c63ef2ae76083a6b4/.github/workflows/ci.yml#L7)

(Sorry, i'm not too familiar with GHA)

--

Also, just for my understanding: Is this related to the other change?

I don't think it is, and that is fine, it is a welcome addition -- just want to make sure i'm folloing along

nickrobinson251 (approved):
I think this looks good! Thanks

Just a couple queries, mostly from my own lack of familiarity with GHA 😊

templates/github/workflows/CI.yml
timholy: Yep! You can see it in https://github.com/JuliaArrays/GetindexArrays.jl/pull/2: the first commit green check has twice as many Actions runs as the second commit green check.

templates/github/workflows/CI.yml
timholy: Unrelated. I'm not sure where I got it from, whether it was reading the actions docs or copying someone else or a PR someone submitted. I like the idea of switching to the one used in Revise :laughing:, I'll update this.

timholy:
One thing I'm not sure of: if you list a `tags` field, does that exclude branches? Or do all branches match unless you specify a specific pattern?

templates/github/workflows/CI.yml
```diff
@@ -1,7 +1,12 @@
 name: CI
 on:
-  - push
-  - pull_request
+  push:
+    <<#BRANCH>>
+    branches:
```
nickrobinson251, line 5: i wonder if we want to swap the order here, so the `branches:` bit is always included?

```suggestion
    branches:
    <<#BRANCH>>
```

timholy: I think you were right to suggest that. From the docs (which answer the question I posed above):
> If you define only tags or only branches, the workflow won't run for events affecting the undefined Git ref.

DilumAluthge:
This is what I use, and it's always worked correctly for me:

```
on:
  pull_request:
    branches:
      - master
  push:
    branches:
      - master
    tags: '*'
```

timholy:
Yeah, the only question is what to do when the user *doesn't* supply `Git(; branch="main")` or whatever.

DilumAluthge:
> Yeah, the only question is what to do when the user _doesn't_ supply `Git(; branch="main")` or whatever.

Put in a reasonable default? We can use Git to read the global Git config and figure out what the default branch name is.

If that fails, just put in either main or master as the default.

timholy:
Actually, it looks like we should just merge #243. Perhaps this should sit here until we know whether we can get that in. 

Let me poke at it for a couple of minutes and see if I can figure it out. My biggest concern is not being able to run the tests locally (see OP), but I see he added `with_clean_gitconfig` which sounds promising as something that might also happen to fix that.

PR #316, opened 2021-10-02
Update "Set default branch explicitly"

timholy:
#243 seems like an important change. This is my crude attempt (without know much about this package or the tools it relies on) to update it to important changes. This most important thing is that #297 didn't fix this PR. I have no idea if this is OK, any chance you could take a glance @mjram0s?

One of the tests that I marked as broken is explained in https://github.com/invenia/PkgTemplates.jl/pull/315#issue-748091329.


timholy:
This is finally passing. I just disabled certain problematic tests for Windows; since this package isn't something I know a lot about, that just seems expedient. I could file an issue as a reminder to turn on Windows again some day. (The error was quite strange, a permissions error on a directory & file created by `mktemp()`. That doesn't seem like it should happen. But I can't easily debug it locally, and for all I know it's a CI-only error.)

nickrobinson251 (approved):

test/fixtures/DocumenterGitLabCI/.gitlab-ci.yml
```diff
@@ -0,0 +1,48 @@
...
+        c, t = get_summary(process_folder())
+        using Printf
+        @printf "Test coverage %.2f%%\n" 100c / t'
+Julia 1.0:
+  image: julia:1.0
+  extends:
+    - .script
+    - .coverage
+Julia 1.5:
+  image: julia:1.5
```
mjram0s, line 25:
```suggestion
Julia 1.6:
  image: julia:1.6
```

timholy:
You did all the hard work, my contribution ended up being a few lines. And I learned a bit (just a bit...) about the internals of an important package for the ecosystem. Thanks again for everything you've done for this package!

PR #332, opened 2021-12-29
GHA: Use julia-actions/cache action

SaschaMann:
See https://github.com/marketplace/actions/cache-julia-artifacts-packages-and-registry

This action caches artifacts and packages and hides the actual config similar to julia-runtest and julia-buildpkg. If a Julia update will require a change to the cache config, people using the template workflow won't have to manually go and fix it everywhere.

The action has options to disable caching of either packages and artifacts and to enable caching of registries, which may be worth considering as an option given to the user via the PkgTemplates interface.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "PkgTemplates"
 uuid = "14b8a8f1-9102-5b29-a752-f990bacb7fe1"
 authors = ["Chris de Graaf", "Invenia Technical Computing Corporation"]
-version = "0.7.24"
+version = "0.7.25"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/templates/github/workflows/CI.yml
+++ b/templates/github/workflows/CI.yml
@@ -47,16 +47,7 @@ jobs:
         with:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
-      - uses: actions/cache@v1
-        env:
-          cache-name: cache-artifacts
-        with:
-          path: ~/.julia/artifacts
-          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
-          restore-keys: |
-            ${{ runner.os }}-test-${{ env.cache-name }}-
-            ${{ runner.os }}-test-
-            ${{ runner.os }}-
+      - uses: julia-actions/cache@v1
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
       <<#HAS_CODECOV>>
--- a/test/fixtures/AllPlugins/.github/workflows/CI.yml
+++ b/test/fixtures/AllPlugins/.github/workflows/CI.yml
@@ -31,16 +31,7 @@ jobs:
         with:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
-      - uses: actions/cache@v1
-        env:
-          cache-name: cache-artifacts
-        with:
-          path: ~/.julia/artifacts
-          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
-          restore-keys: |
-            ${{ runner.os }}-test-${{ env.cache-name }}-
-            ${{ runner.os }}-test-
-            ${{ runner.os }}-
+      - uses: julia-actions/cache@v1
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
       - uses: julia-actions/julia-processcoverage@v1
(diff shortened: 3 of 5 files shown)


SaschaMann:
I searched for `actions/cache@v1` and replaced all occurrences but I'm not familiar enough with PkgTemplates to tell whether that's everything that needs to be changed.

nickrobinson251 (approved):
Thanks! I'd be happy to merge this -- just a couple quick questions before i do (since i wasn't familiar with the existing `cache` action; also FYI @christopher-dG who i think added the original action in https://github.com/invenia/PkgTemplates.jl/pull/173)

From the page you linked, it says the benefit of changing `actions/cache@v1` -> `julia-actions/cache@v1` is that it'll be faster for our use-case; are there any downsides to changing?

Also, can you say why the existing arguments (`cache-name`, `key`, etc.) are no longer needed (i couldn't see why this was used before and not now by just looking at the linked page).

> The action has options to disable caching of either packages and artifacts and to enable caching of registries, which may be worth considering as an option given to the user via the PkgTemplates interface.

Feel encouraged to open an issue to track this idea (or make a PR for it), if it is something you think would be useful.

> that's everything that needs to be changed

yep, looks good.

SaschaMann:
> From the page you linked, it says the benefit of changing actions/cache@v1 -> julia-actions/cache@v1 is that it'll be faster for our use-case; are there any downsides to changing?

It won't be faster than `actions/cache@v1`. It's effectively a wrapper around it to simplify the workflow. You can see which commands it runs here: https://github.com/julia-actions/cache/blob/main/action.yml - There is some extra logic because it also caches packages whereas the current workflow only caches artifacts. The main upside is not the speed but the ease of maintainability of the workflow as mentioned in the OP of this PR.

The major downsides that I can see:

- It hides what's actually being run so there's an extra layer of "obfuscation". Since all the other Julia actions (`julia-runtest` etc) do the same, I think this is acceptable. 
- Since it calls `actions/cache` internally and pins that action to a commit SHA, updates to `actions/cache` won't be available immediately upon release but will require an update to `julia-actions/cache`. For the user of the workflow PkgTemplate generates, this won't result in any additional work but it's technically a downside.

> Also, can you say why the existing arguments (cache-name, key, etc.) are no longer needed (i couldn't see why this was used before and not now by just looking at the linked page).

The key is used to specify which cache should be accessed. For example if you have different dependencies on a branch, e.g. because a PR introduces a new dependency, it will use a separate cache for that branch compared to what's running on `main`. `restore-key`s are fallback keys if there's no exact match because a partial match might still result in a speed up over not using any cache at all.

The `cache-name` is used as part of the key. I'm not sure why `cache-name` was used by default before either (I can't really think of a use case) but the current behaviour can be manually replicated by setting the `cache-name` input if someone does want to change the name.

The keys and default cache-name are set internally in the new version and are largely identical with the current version, though some fallback keys that weren't particularly useful have been removed.

Before:

```yaml
        with:
          path: ~/.julia/artifacts
          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
          restore-keys: |
            ${{ runner.os }}-test-${{ env.cache-name }}-
            ${{ runner.os }}-test-
            ${{ runner.os }}-
```

After (internally):

```yaml
        key: ${{ runner.os }}-test-${{ inputs.cache-name }}-${{ hashFiles('**/Project.toml') }}
        restore-keys: |
          ${{ runner.os }}-test-${{ inputs.cache-name }}-
```

> Feel encouraged to open an issue to track this idea (or make a PR for it), if it is something you think would be useful.

I'm not sure if it's useful tbh. I brought it up so that it's documented as a possibility in case someone requests it. I can open an issue but don't think a PR is worth it until someone asks for it.

nickrobinson251:
Thanks for the expanations! And all sounds good to me.

I wonder if the CI for this repo itself should also switch to `julia-actions/cache`? cc @fchorney who last touched this config in https://github.com/invenia/PkgTemplates.jl/pull/263. But this can be it's own discussion/issue/PR -- doesn't have to be added to this PR.

If you bump the version number, i will make a release with this change as soon as this is merged. Thanks, again!

SaschaMann:
> I wonder if the CI for this repo itself should also switch to julia-actions/cache?

It won't hurt for the same reasons as above, but I don't think there's an immediate need to switch over since the behaviour is similar enough.

As a side note, if you're changing to the new cache action, I'd also recommend switching to `@v1` instead of `@latest` for `julia-buildpkg` and `julia-runtest`. While it's unlikely there will be breaking changes to those actions, it's not impossible and using `@latest` could cause issues in that case. See #237.

> If you bump the version number, i will make a release with this change as soon as this is merged. Thanks, again!

Is this a minor or patch change? I'm not sure what counts as part of the interface in this package.

nickrobinson251:
Patch/non-breaking bump please :)

rikhuijzer:
> It won't be faster than actions/cache@v1. 

Actually, it will. The old one did not cache `~/.julia/packages` and the new one does.

Issue #270, opened 2021-02-04
Badge Plugins for JuliaHub badges (Dependents, PkgEval, Version)


nickrobinson251:
I noticed the `Dependents` badge on the [Arrow.jl](https://github.com/JuliaData/Arrow.jl/) repo, and thought it was kinda cool. 

Turns out this is through [JuliaHub](https://docs.juliahub.com/), and they provide `Version` and `PkgEval` badges too, see e.g. https://discourse.julialang.org/t/package-badges-on-juliahub/44635

These could easily be added as [`BadgePlugin`](https://invenia.github.io/PkgTemplates.jl/stable/developer/#PkgTemplates.BadgePlugin)s _if_ we new how the URL is generated e.g.
```julia
struct DependentsBadge <: BadgePlugin end

function badges(::DependentsBadge)
    return Badge(
        "Dependent Packages",
        "https://juliahub.com/docs/{{{PKG}}}/deps.svg",
        "https://juliahub.com/ui/Packages/{{{PKG}}}/QnF3w",  # <- need to know what these 5 "random" chars will be
    )
end
```
I suppose we'd also need a `view` method, like `view(::DependentsBadge, ::Template, pkg::AbstractString) = Dict("PKG" => pkg, "RANDOM_CHARS" => "QnF3w")`

So i don't know if this would be possible, but maybe someone who works on JuliaHub knows :)

nickrobinson251:
maybe @aviks knows?

pfitzseb:
You can get the package slug with `Base.package_slug(pkg_uuid, 5)`, so e.g. 
```
julia> Base.package_slug(UUID("14b8a8f1-9102-5b29-a752-f990bacb7fe1"), 5)
"IGiQL"
```
for this package: [![deps](https://juliahub.com/docs/PkgTemplates/deps.svg)](https://juliahub.com/ui/Packages/PkgTemplates/IGiQL?t=2)

nickrobinson251:
thanks! hmm, okay. I guess we could get the UUID from the Project.toml of the newly created repo (if the new repo is the active env), with something like
```julia
uuid_line = readlines(Base.active_project())[2];
uuid = only(match(r"\"(.*?)\"", uuid_line).captures);
slug = Base.package_slug(UUID(uuid), 5)
```
maybe/hopefully someone can come up with something nicer

christopher-dG:
[...]

PR #334, opened 2022-01-05
add PkgEval badge

mileslucas:
- add PkgEval badge
- add tests for PkgEvalBadge

An example of this badge can be seen in this repo- https://github.com/JuliaPhysics/Mueller.jl


src/plugins/readme.jl
```diff
@@ -48,6 +48,7 @@ function view(p::Readme, t::Template, pkg::AbstractString)
         "HAS_CITATION" => hasplugin(t, Citation) && getplugin(t, Citation).readme,
         "HAS_INLINE_BADGES" => !isempty(strings) && p.inline_badges,
         "PKG" => pkg,
+        "PKG1" => first(pkg) # for PkgEval badge, need first character
```
mileslucas, line 51: Happy to use a more descriptive variable name if you can think of one.

test/show.jl
```diff
@@ -19,7 +19,7 @@ end
               file: "$(joinpath(TEMPLATES_DIR, "README.md"))"
               destination: "README.md"
               inline_badges: false
-              badge_order: DataType[Documenter{GitHubActions}, Documenter{GitLabCI}, Documenter{TravisCI}, GitHubActions, GitLabCI, TravisCI, AppVeyor, DroneCI, CirrusCI, Codecov, Coveralls, BlueStyleBadge, ColPracBadge]
+              badge_order: DataType[Documenter{GitHubActions}, Documenter{GitLabCI}, Documenter{TravisCI}, GitHubActions, GitLabCI, TravisCI, AppVeyor, DroneCI, CirrusCI, Codecov, Coveralls, BlueStyleBadge, ColPracBadge, PkgEvalBadge]
```
mileslucas, line 22: I have no clue why this is failing. It fails locally, too.

If I just
```julia
julia> using PkgTemplates

julia> Readme()
Readme:
  file: "~/dev/julia/PkgTemplates/templates/README.md"
  destination: "README.md"
  inline_badges: false
  badge_order: DataType[Documenter{GitHubActions}, Documenter{GitLabCI}, Documenter{TravisCI}, GitHubActions, GitLabCI, TravisCI, AppVeyor, DroneCI, CirrusCI, Codecov, Coveralls, BlueStyleBadge, ColPracBadge, PkgEvalBadge]
  badge_off: DataType[]
```
the `PkgEvalBadge` shows up just fine. It's like somehow the test suite, both locally and on CI, is ignoring the changes I've made to the code, or else that's occurring in the `show` pipeline somehow.

nickrobinson251: looks like it's the next testset below that is failing. think the `badge_order: ...` line on L59 needs updating too

src/plugins/readme.jl
nickrobinson251: i think this is fine with the comment :) i suppose "PKG_INITIAL" is an alternative

src/plugins/badges.jl
```diff
@@ -34,3 +34,18 @@ function badges(::ColPracBadge)
         "https://github.com/SciML/ColPrac",
     )
 end
+
+"""
+    PkgEvalBadge()
+
+Adds a [`PkgEval`](https://github.com/JuliaCI/PkgEval.jl) badge to the [`Readme`](@ref) file.
```
nickrobinson251, line 41: Since the PkgEval.jl link doesn't explain mention the badge, perhaps we should link to https://github.com/JuliaCI/NanosoldierReports#badges too somehow?

Perhaps either:
```suggestion
Adds a [`PkgEval`](https://github.com/JuliaCI/PkgEval.jl) [badge](https://github.com/JuliaCI/NanosoldierReports#badges) to the [`Readme`](@ref) file.
```
or maybe better:
```suggestion
Adds a [`PkgEval` badge](https://github.com/JuliaCI/NanosoldierReports#pkgeval-reports) to the [`Readme`](@ref) file.
```

nickrobinson251 (approved):
Thanks! Looks great!

Just one more test to update i think, plus a small suggestion :)

test/show.jl
mileslucas: Oh, good catch! That fixed it

mileslucas:
It should be ready to go! Thanks for the review.

src/plugins/readme.jl
nickrobinson251: Apologies for missing this initially, but we need to add a `view` method for the `PkgEvalBadge` rather than extending the `Readme` method

I think this should be sufficient: 
```jl
view(::PkgEvalBadge, t::Template, pkg::AbstractString) = Dict("PKG1" => first(pkg), "PKG" => pkg)
```

And in order to make sure this works correctly, we should add `PkgEvalBadge()` to the "Wacky Options" reference tests in `test/reference.jl`:
- you just need to add it to the `test/` code, then if you run the tests using Julia v1.5 (only this version runs the reference tests)
- then it should give you the option to automatically update the actually reference files. 
- Of course, we then need to check the updated README.md file that gets generated has the expected badge with correct URLs.

(also xref #270)

mileslucas: Incorporated.

Note: I actually ran into issues running this in docker-
```
$ docker run --rm -ti -v $(pwd):/code julia:1.5 bash
# cd code
# julia --project=@.
julia>]
(PkgTemplates) pkg> test
ERROR: AssertionError: sourcepath !== nothing
```

So, I just changed the expected README file manually.

nickrobinson251: ah, i think `ERROR: AssertionError: sourcepath !== nothing` is a cryptic Pkg.jl error, related to trying to use v1.5 withan env with a Manifest.toml generated in with Julia v1.6+  ...or something like that. Anyway, sorry it was a totally smooth process! But the manual change LGTM and CI is happy

Thanks again!

PR #335, opened 2022-01-06
Generate Project.toml with Pkg's custom key order, and update reference tests to v1.7

### LINKED ISSUE #298: Update reference tests for Julia 1.6
`reference.jl` is only tested on Julia 1.5 (https://github.com/invenia/PkgTemplates.jl/blob/master/test/runtests.jl#L67) and should be updated to be compatible with Julia 1.6

tkf:
In Julia 1.7, Project.toml generated by PkgTemplates is not compatible with Pkg's sort order.

```
(@v1.7) pkg> activate --temp
  Activating new project at `/tmp/jl_ZsryYH`

julia> using PkgTemplates
 │ Package PkgTemplates not found, but a package named PkgTemplates is available from a registry.
 │ Install package?
 │   (jl_ZsryYH) pkg> add PkgTemplates
 └ (y/n) [y]:
...

(jl_ZsryYH) pkg> st
      Status `/tmp/jl_ZsryYH/Project.toml`
  [14b8a8f1] PkgTemplates v0.7.25

julia> Template(user = "tkf")("ExamplePackage")
[ Info: Running prehooks
...
[ Info: New package is at /home/arakaki/.julia/dev/ExamplePackage

shell> cd ~/.julia/dev/ExamplePackage
/home/arakaki/.julia/dev/ExamplePackage

shell> cat Project.toml
name = "ExamplePackage"
uuid = "db43d3c3-1afb-45d2-b722-e2cdd05cc17f"
authors = ["Takafumi Arakaki <aka.tkf@gmail.com> and contributors"]
version = "0.1.0"

[extras]
Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"

[compat]
julia = "1"

[targets]
test = ["Test"]
```

The order of the tables is different from how Pkg sorts them. If I update it with Pkg:

```
(jl_ZsryYH) pkg> activate .
  Activating project at `~/.julia/dev/ExamplePackage`

(ExamplePackage) pkg> add UUIDs
...
```

Then `git diff` shows me

```diff
diff --git a/Project.toml b/Project.toml
index 3d3b37e..922c5e8 100644
--- a/Project.toml
+++ b/Project.toml
@@ -3,11 +3,14 @@ uuid = "db43d3c3-1afb-45d2-b722-e2cdd05cc17f"
 authors = ["Takafumi Arakaki <aka.tkf@gmail.com> and contributors"]
 version = "0.1.0"

-[extras]
-Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+[deps]
+UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"

 [compat]
 julia = "1"

+[extras]
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+
 [targets]
 test = ["Test"]
```

This PR solves this issue with a custom `write_project` with the sort key taken form Pkg.

It also bumps the Julia version used for the reference test to check the bug this PR fixes. To make it work, I added a mocking for `uuid4`.

close #298

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -40,6 +40,10 @@ jobs:
           - os: ubuntu-latest
             version: 1.5
             arch: x64
+          # Add 1.7.1 to run the reference tests
+          - os: ubuntu-latest
+            version: 1.7.1
+            arch: x64
     steps:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -36,10 +36,6 @@ jobs:
           - os: windows-latest
             arch: x86
         include:
-          # Add a 1.5 job because that's what Invenia actually uses
-          - os: ubuntu-latest
-            version: 1.5
-            arch: x64
           # Add 1.7.1 to run the reference tests
           - os: ubuntu-latest
             version: 1.7.1
--- a/src/plugins/project_file.jl
+++ b/src/plugins/project_file.jl
@@ -26,9 +26,11 @@ end
 
 # Taken from:
 # https://github.com/JuliaLang/Pkg.jl/blob/v1.7.0/src/project.jl#L175-L177
-_project_key_order = ["name", "uuid", "keywords", "license", "desc", "deps", "compat"]
-project_key_order(key::String) =
-    something(findfirst(x -> x == key, _project_key_order), length(_project_key_order) + 1)
+
+function project_key_order(key::String)
+    _project_key_order = ["name", "uuid", "keywords", "license", "desc", "deps", "compat"]
+    return something(findfirst(x -> x == key, _project_key_order), length(_project_key_order) + 1)
+end
 
 write_project(path::AbstractString, dict) =
     open(io -> write_project(io, dict), path; write = true)
--- a/src/plugins/project_file.jl
+++ b/src/plugins/project_file.jl
@@ -16,7 +16,7 @@ priority(::ProjectFile, ::typeof(hook)) = typemax(Int) - 5
 function hook(p::ProjectFile, t::Template, pkg_dir::AbstractString)
     toml = Dict(
         "name" => basename(pkg_dir),
-        "uuid" => string(uuid4()),
+        "uuid" => string(@mock uuid4()),
         "authors" => t.authors,
         "version" => string(p.version),
         "compat" => Dict("julia" => compat_version(t.julia)),
--- a/src/plugins/project_file.jl
+++ b/src/plugins/project_file.jl
@@ -21,8 +21,19 @@ function hook(p::ProjectFile, t::Template, pkg_dir::AbstractString)
         "version" => string(p.version),
         "compat" => Dict("julia" => compat_version(t.julia)),
     )
-    open(io -> TOML.print(io, toml), joinpath(pkg_dir, "Project.toml"), "w")
+    write_project(joinpath(pkg_dir, "Project.toml"), toml)
 end
+
+# Taken from:
+# https://github.com/JuliaLang/Pkg.jl/blob/2341a067973467949653928ab9f49d98cae6057e/src/project.jl#L175-L177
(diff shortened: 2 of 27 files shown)


.github/workflows/CI.yml
```diff
@@ -40,6 +40,10 @@ jobs:
           - os: ubuntu-latest
             version: 1.5
             arch: x64
+          # Add 1.7.1 to run the reference tests
```
nickrobinson251, line 43: We can ditch the `Add a 1.5 job ...` code above now too. i see no reason to stick to test on whatever version Invenia actually uses in Production, which is v1.6 now anyway, as we don't use PkgTemplates.jl in Prod. (GitHub won't let me make an edit suggestion on uneditted lines unfortunately).

src/plugins/project_file.jl
```diff
@@ -16,14 +16,25 @@ priority(::ProjectFile, ::typeof(hook)) = typemax(Int) - 5
...
         "authors" => t.authors,
         "version" => string(p.version),
         "compat" => Dict("julia" => compat_version(t.julia)),
     )
-    open(io -> TOML.print(io, toml), joinpath(pkg_dir, "Project.toml"), "w")
+    write_project(joinpath(pkg_dir, "Project.toml"), toml)
 end
 
+# Taken from:
+# https://github.com/JuliaLang/Pkg.jl/blob/2341a067973467949653928ab9f49d98cae6057e/src/project.jl#L175-L177
```
nickrobinson251, line 28: for a little more context, let's link to a specific tag version of the code:
```suggestion
# https://github.com/JuliaLang/Pkg.jl/blob/v1.7.0/src/project.jl
```

src/plugins/project_file.jl
```diff
@@ -16,14 +16,25 @@ priority(::ProjectFile, ::typeof(hook)) = typemax(Int) - 5
...
+_project_key_order = ["name", "uuid", "keywords", "license", "desc", "deps", "compat"]
+project_key_order(key::String) =
+    something(findfirst(x -> x == key, _project_key_order), length(_project_key_order) + 1)
```
nickrobinson251, line 31: i know this is a direct copy-paste of the linked code, but using a non-const global variable does seem very odd, so i think i prefer we change it for this repo:
```suggestion

function project_key_order(key::String)
    _project_key_order = ["name", "uuid", "keywords", "license", "desc", "deps", "compat"]
    return something(findfirst(x -> x == key, _project_key_order), length(_project_key_order) + 1)
end
```

nickrobinson251 (approved):
Thanks a lot!

A couple small suggestions that i'd appreciate.

> It also bumps the Julia version used for the reference test to check the bug this PR fixes. 

This will close #298, which has been on the wish-list a long while, so i'm very grateful for you doing it!

> To make it work, I added a mocking for uuid4.

Just to check my understanding: this mocking is needed as the output of `uuid4()` is different in v1.6+ (post https://github.com/JuliaLang/julia/pull/35872)?

tkf:
Thanks for the review!

> This will close #298

FYI I added `close #298` to the OP to auto-close it. But feel free to remove it if you manage it differently.

> Just to check my understanding: this mocking is needed as the output of `uuid4()` is different in v1.6+ (post [JuliaLang/julia#35872](https://github.com/JuliaLang/julia/pull/35872))?

Oooh, this explains why. I was a bit puzzled why, but I agree that it explains it. (I initially speculated that the test may actually not be running because that 1.5 was not in the CI matrix before realizing the explicit `include`)

nickrobinson251:
hmm, it seems deeleting the 1.5 job from the GHA workflow file might not be sufficient, as it seems the repo is somehow still expecting a 1.5 job to run and pass... I wonder if this is something that needs changing in the repo settings (to which i don't have access)  

Perhaps @christopher-dG knows? And/or perhaps @oxinabox might be able to bump my permissions to allow me to access repo settiing to investigate?

christopher-dG:
I've replaced the 1.5 required check in the repo settings with a 1.7 one :)

nickrobinson251:
Thanks!

PR #337, opened 2022-01-14
changed link for Travis CI badge

YanniPapandreou:
Badge link for Travis CI was pointing to "https://travis-ci.com/{{{USER}}}/{{{PKG}}}.jl" which does not seem to be the correct link. Changed this to "https://app.travis-ci.com/{{{USER}}}/{{{PKG}}}.jl" which seems to work. :)

close https://github.com/invenia/PkgTemplates.jl/issues/338


nickrobinson251:
Thanks! Can you find any indo (eg. a blog post by TravisCI) that indicates that this changed and this update URL is the right one?

nickrobinson251:
For tests to pass, you will need to update the reference tests to have the updated URL. You can do this by running in tests locally using Julia 1.7, and you should see the reference tests fail _and be giving the option to have them automatically updated for you_ (https://invenia.github.io/PkgTemplates.jl/dev/developer/#Updating-Reference-Tests-and-Fixtures-1) -- let me know if you run into any issues with that!

YanniPapandreou:
> Thanks! Can you find any info (eg. a blog post by TravisCI) that indicates that this changed and this update URL is the right one?

When I used the original link it took me to an "error 404 page not found" even though on my Travis account I could find the repo being checked. The link was the proposed change. Not sure however if this might have to do with my OS?

YanniPapandreou:
> For tests to pass, you will need to update the reference tests to have the updated URL. You can do this by running in tests locally using Julia 1.7, and you should see the reference tests fail _and be given the option to have them automatically updated for you_ (https://invenia.github.io/PkgTemplates.jl/dev/developer/#Updating-Reference-Tests-and-Fixtures-1) -- let me know if you run into any issues with that!

So I think the same test fails even on the free package without me making my change. However, I have tried starting my REPL, then accessing the Pkg REPL doing `dev PkgTemplates` making my changes and then finally in the Pkg REPL doing `test PkgTemplates`. The tests run and the reference test fails like you say, however, I do not seem to have an option to automatically change the URL. Is there another why I should be running the tests? I am not understanding the link you sent sorry, I am very new to all of this.

gdalle (approved):

PR #340, opened 2022-02-08
Add slash to the end of docs URLs

bicycle1885:
I found that GitHub Pages redirects an URL without a slash suffix to a corresponding URL with a slash. For example, https://invenia.github.io/PkgTemplates.jl/dev is redirected to https://invenia.github.io/PkgTemplates.jl/dev/. It is expected but I think this redirection is just a waste of time. For example, on my Firefox, 155 ms was wasted for redirection (see the following screenshot).

image

To make matters worse, redirection seems to be disabled when pages are hosted on a private repository. Thus, the default URLs respond a 404 error code. I didn't confirm reproducibility of this so my observation may be just a mistake.


nickrobinson251 (approved):
thanks

PR #341, opened 2022-02-12
GitHub Actions CI template: do not skip intermediate builds on the `main`, `master`, or `release-*` branches

DilumAluthge:
Follow-up to #325

### DIFF
--- a/templates/github/workflows/CI.yml
+++ b/templates/github/workflows/CI.yml
@@ -9,9 +9,9 @@ on:
   pull_request:
   workflow_dispatch:
 concurrency:
-  # Skip intermediate builds: always.
-  # Cancel intermediate builds: only if it is a pull request build.
-  group: ${{ github.workflow }}-${{ github.ref }}
+  # Skip intermediate builds: all builds except for builds on the `main`, `master`, or `release-*` branches
+  # Cancel intermediate builds: only pull request builds
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.ref != 'refs/heads/main' || github.ref != 'refs/heads/master' || startsWith(github.ref, 'refs/heads/release-') || github.run_number }}
   cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
 jobs:
   test:


DilumAluthge:
Bump @christopher-dG @nickrobinson251

nickrobinson251 (commented):
Seems sensible to me!

Looks like the reference tests need updating (they only run on Julia v1.7.1 which is why that CI job fails), then LGTM

Also please bump the version and I'll release straight away

Thanks!

templates/github/workflows/CI.yml
```diff
@@ -8,9 +8,9 @@ on:
     tags: '*'
   pull_request:
 concurrency:
-  # Skip intermediate builds: always.
-  # Cancel intermediate builds: only if it is a pull request build.
-  group: ${{ github.workflow }}-${{ github.ref }}
+  # Skip intermediate builds: all builds except for builds on the `main`, `master`, or `release-*` branches
+  # Cancel intermediate builds: only pull request builds
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.ref != 'refs/heads/main' || github.ref != 'refs/heads/master' || startsWith(github.ref, 'refs/heads/release-') || github.run_number }}
```
gdalle, line 13: can you explain a little more what this does? I read https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency but I'm still a bit confused

DilumAluthge: So, roughly speaking, if two separate builds have the same concurrency `group` value, then the second build will have to wait until the first build has completed. Additionally, if three builds all have the same concurrency `group` value, then while the first build is happening, the second and third build will have to wait, but once the first build has completed, the second build will be completely skipped, and the third build will start.

This is good for PRs, but usually it is undesirable to not run CI on every commit pushed to `main` or `master`. Therefore, for commits on `main`, `master`, and `release-*`, we create a unique concurrency `group` value. This ensures that two different commits on `main` never have the same concurrency `group` value, and thus builds on `main` are never skipped.

For commits on PRs, two separate commits on the same PR will have the same concurrency `group` value. Therefore, if multiple commits are pushed to a PR, we can skip CI on the intermediate commits, and only run CI on the latest commit of the PR, since that is the code that we actually care about.

DilumAluthge:
@gdalle @nickrobinson251 How would I go about fixing the CI failures here?

PR #346, opened 2022-04-05
Interactive mode bugfix for a single customizable option

rossviljoen:
Fixes a bug where an error would be thrown if there is only a single customisable user option when using interactive mode. This PR implements what I believe was the intended behaviour.


nickrobinson251:
thanks, @rossviljoen 

what was the incorrect behaviour you observed? and do you think it might be possible to add a test for it (possibly in `test/interactive.jl`)?

src/template.jl
```diff
@@ -207,7 +207,7 @@ function interactive(::Type{Template}; kwargs...)
         println("Template keywords to customize:")
         menu = MultiSelectMenu(map(string, customizable); pagesize=length(customizable))
         customize = customizable[sort!(collect(request(menu)))]
-        just_one && lastindex(customizable) in customize && return Template(; kwargs...)
+        just_one && last(customizable) in customize && return Template(; kwargs...)
```
gdalle, line 210: this seems reasonable given that we want to test if the `customizable` list is just `[:none]`

gdalle (approved):

PR #347, opened 2022-05-16
Set `permissions: contents: write`

rikhuijzer:
Setting `permissions: contents: write` is required in new repositories for the `GITHUB_TOKEN` to work. See also https://github.com/JuliaDocs/Documenter.jl/pull/1819 and https://github.com/julia-actions/julia-docdeploy/pull/21

Without it, people might get an error like
```
fatal: unable to access 'https://github.com/rikhuijzer/MyProject.jl.git/': The requested URL returned error: 403
```

Please double-check whether I haven't missed a place in PkgTemplates.jl where this setting should be set too

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -6,10 +6,13 @@ on:
 jobs:
   CompatHelper:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     steps:
       - name: Pkg.add("CompatHelper")
         run: julia -e 'using Pkg; Pkg.add("CompatHelper")'
       - name: CompatHelper.main()
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          COMPATHELPER_PRIV: ${{ secrets.DOCUMENTER_KEY }}
         run: julia -e 'using CompatHelper; CompatHelper.main()'
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -6,6 +6,7 @@ on:
 jobs:
   CompatHelper:
     runs-on: ubuntu-latest
+    permissions: write-all
     steps:
       - name: Pkg.add("CompatHelper")
         run: julia -e 'using Pkg; Pkg.add("CompatHelper")'
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -6,12 +6,10 @@ on:
 jobs:
   CompatHelper:
     runs-on: ubuntu-latest
-    permissions: write-all
     steps:
       - name: Pkg.add("CompatHelper")
         run: julia -e 'using Pkg; Pkg.add("CompatHelper")'
       - name: CompatHelper.main()
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          COMPATHELPER_PRIV: ${{ secrets.DOCUMENTER_KEY }}
         run: julia -e 'using CompatHelper; CompatHelper.main()'
--- a/templates/github/workflows/register.yml
+++ b/templates/github/workflows/register.yml
@@ -8,6 +8,7 @@ on:
 jobs:
   register:
     runs-on: ubuntu-latest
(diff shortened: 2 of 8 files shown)


nickrobinson251:
Thanks for the contribution, @rikhuijzer 

Looks like a few more files in `test/fixtures/` need to the same update for reference tests to pass.

More generally, are there more restricted permissions we could set then `write-all` that would enable what we want? 

PkgTemplates is fairly widely used to whatever permissions we set here can end up being set across many packages, so i'd be keen to set the minimal required :)

DilumAluthge:
> More generally, are there more restricted permissions we could set then `write-all` that would enable what we want?

Yeah, I wojld recommend

rikhuijzer:
> More generally, are there more restricted permissions we could set then write-all that would enable what we want?

Yes. Well spotted. Would you agree that 
```
    permissions:
      contents: write
```
is good? I've tested it at https://github.com/rikhuijzer/DiscountedCashFlows.jl/commits/main commits 4f85500df75d6efc78e999ed78e201a17f74216d and d19106abd4de732fdecc8d64aa1e176058dc855c.

nickrobinson251:
Yeah, `permissions: contents: write` sounds good to me!

I think you'll also need to rebase this on `master`, as we just updated the reference tests to run on v1.7.2 (#349).
If you then run tests locally with v1.7.2 it should offer the option to update any reference test files that need updating to match this change.

Finally, if you bump the Project.toml version to the next non-breaking version, i'll make a release with this change as soon as it's merged

Thanks again!

templates/github/workflows/register.yml
```diff
@@ -8,6 +8,8 @@ on:
 jobs:
   register:
     runs-on: ubuntu-latest
+    permissions:
+        contents: write
```
rikhuijzer, line 12: I'm not sure whether this is enough for the register action. The GitHub Actions scope elements are very poorly documented (e.g., https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/)

.github/workflows/CompatHelper.yml
```diff
@@ -6,6 +6,8 @@ on:
 jobs:
   CompatHelper:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
```
rikhuijzer, line 10:
```suggestion
    permissions:
      contents: write
```
Thanks to the unclear GitHub docs, I'm usure whether 
```yaml
permissions
   contents: write
   pull-requests: write
```
without the compathelper_priv key would be enough

rikhuijzer:
> I think you'll also need to rebase this on master, as we just updated the reference tests to run on v1.7.2 (https://github.com/invenia/PkgTemplates.jl/pull/349).
If you then run tests locally with v1.7.2 it should offer the option to update any reference test files that need updating to match this change.

> Finally, if you bump the Project.toml version to the next non-breaking version, i'll make a release with this change as soon as it's merged

Great. Thanks for the instructions.

I won't have time to continue working on this. If you want to get it merged quickly, then feel free to make some further adjustements

nickrobinson251:
thanks for your work, @rikhuijzer 

hopefully i can get this over the line from here. i must admit Github Actions is not something i'm too familiar with. I wonder if @DilumAluthge might be able to give this PR a review?

.github/workflows/CompatHelper.yml
rikhuijzer, line 10:
```suggestion
    permissions:
      contents: write
```
It could be that we can drop COMPATHELPER_PRIV if we set
```yaml
permissions:
    contents: write
    pull-requests: write
```
I've tested this and it didn't work. This could be due to:

1. GITHUB_TOKEN not having enough permissions open a PR
2. CompatHelper not reading the GITHUB_TOKEN even when permissions are set

DilumAluthge: You need the `COMPATHELPER_PRIV` to get GitHub Actions CI to automatically trigger on PRs opened by CompatHelper.

rikhuijzer: Yes. Sounds like it

nickrobinson251 (approved):
Thanks again!

PR #350, opened 2022-05-25
Allow setting Documenter `edit_link`

### LINKED ISSUE #348: "Edit on GitHub" link in Documenter.jl
Hi, and thanks for this amazing package!
At the moment, repos set up with PkgTemplates.jl work perfectly with Documenter.jl, except for the "Edit on GitHub" link, which always points to `master` instead of the chosen branch. The reason behind that is outlined here, as well as an easy fix: https://github.com/JuliaDocs/Documenter.jl/issues/1375
Would it be possible to add the argument `edit_branch = ...` to `Documenter.HTML` inside `docs/make.jl` when building a template?

nickrobinson251:
- close #348 
- Adds support for `edit_link` which was added in Documenter v0.24 (previously there was `edit_branch`) https://github.com/JuliaDocs/Documenter.jl/pull/1173
- since `edit_link=nothing` has a meaning in Documenter, we use `edit_link=:devbranch` to mean "do the obvious thing" which i think it is uncontroversial to say is "point at the main development branch" which we take from the `devbranch=...` keyword (which in turn fallsback to `default_branch(::Template)`)
  - also adds support for users passing Symbols, both so we can use `:devbranch` as the default and because Documenter allows Symbols (specifically [the Symbol `:commit` has a special meaning](https://github.com/JuliaDocs/Documenter.jl/blob/8986c4d6e1b58ffc009f8b3733fc57520f4a73a9/src/Writers/HTMLWriter.jl#L307-L311))

cc @gdalle

### DIFF
--- a/src/interactive.jl
+++ b/src/interactive.jl
@@ -93,8 +93,10 @@ function convert_input(P::Type, ::Type{Union{T, Nothing}}, s::AbstractString) wh
 end
 
 function convert_input(P::Type, ::Type{Union{T, Symbol, Nothing}}, s::AbstractString) where T
-    return if startswith(s, ":") || startswith(s, "Symbol(\"")
-        Symbol(s)
+    # Assume inputs starting with ':' char are intended as Symbols, if a plugin accept symbols.
+    # i.e. assume the set of valid Symbols the plugin expects can be spelt starting with ':'.
+    return if startswith(s, ":")
+        Symbol(s[2:end])  # skip ':'
     else
         convert_input(P, Union{T,Nothing}, s)
     end
--- a/src/interactive.jl
+++ b/src/interactive.jl
@@ -96,7 +96,7 @@ function convert_input(P::Type, ::Type{Union{T, Symbol, Nothing}}, s::AbstractSt
     # Assume inputs starting with ':' char are intended as Symbols, if a plugin accept symbols.
     # i.e. assume the set of valid Symbols the plugin expects can be spelt starting with ':'.
     return if startswith(s, ":")
-        Symbol(s[2:end])  # skip ':'
+        Symbol(chop(s, head=1, tail=0))  # remove ':'
     else
         convert_input(P, Union{T,Nothing}, s)
     end
--- a/src/interactive.jl
+++ b/src/interactive.jl
@@ -92,6 +92,14 @@ function convert_input(P::Type, ::Type{Union{T, Nothing}}, s::AbstractString) wh
     return s == "nothing" ? nothing : convert_input(P, T, s)
 end
 
+function convert_input(P::Type, ::Type{Union{T, Symbol, Nothing}}, s::AbstractString) where T
+    return if startswith(s, ":") || startswith(s, "Symbol(\"")
+        Symbol(s)
+    else
+        convert_input(P, Union{T,Nothing}, s)
+    end
+end
+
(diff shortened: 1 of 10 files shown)


nickrobinson251:
need #349 to go in so reference tests run on the newly expected version

nickrobinson251:
rebased now #349 is merged

src/interactive.jl
```diff
@@ -92,6 +92,14 @@ function convert_input(P::Type, ::Type{Union{T, Nothing}}, s::AbstractString) wh
     return s == "nothing" ? nothing : convert_input(P, T, s)
 end
 
+function convert_input(P::Type, ::Type{Union{T, Symbol, Nothing}}, s::AbstractString) where T
+    return if startswith(s, ":") || startswith(s, "Symbol(\"")
+        Symbol(s)
```
oxinabox, line 97: This seems like it won't work right.
It will make `Symbol(":foo")` or `Symbol("Symbol(\"foo\")"`
but don't we want `Symbol("foo")`?

nickrobinson251: ah, yes, we need to chop off the `:` prefix - thanks

we don't actually need to support `Symbol("foo bar")`, so we can drop that clause

nickrobinson251:
Issue about Documenter matching this behaviour of defaulting `edit_link` to the default branch (as with `devbranch`): 
https://github.com/JuliaDocs/Documenter.jl/issues/1827

src/interactive.jl
```diff
@@ -93,8 +93,10 @@ function convert_input(P::Type, ::Type{Union{T, Nothing}}, s::AbstractString) wh
 end
 
 function convert_input(P::Type, ::Type{Union{T, Symbol, Nothing}}, s::AbstractString) where T
-    return if startswith(s, ":") || startswith(s, "Symbol(\"")
-        Symbol(s)
+    # Assume inputs starting with ':' char are intended as Symbols, if a plugin accept symbols.
+    # i.e. assume the set of valid Symbols the plugin expects can be spelt starting with ':'.
+    return if startswith(s, ":")
+        Symbol(s[2:end])  # skip ':'
```
oxinabox, line 99: Want to see something you will hate?
```suggestion
    return if (m=match(s, ":(.*)")) !== nothing
        Symbol(only(m))
```

oxinabox (approved):

src/interactive.jl
nickrobinson251: in what way would this be better? i suppose it avoids assuming `s[2:end]` will work (i.e. we could have the string `":"`) ...maybe we should switch to `chop` to avoid that? is there another rasons a regex would be better?

```suggestion
    return if startswith(s, ":")
        chop(s, head=1, tail=0)  # skip ':'
```

nickrobinson251: (we should drop v1.0 but that's another task)

nickrobinson251: i suppose if we wanted the string `":"` to git the string path below, not the symbol path, we could do `(m=match(s, r":(.{1,})")) !== nothing` and probably a regex is best for that behaviour? i am curious about why a regex might be preferred 😊

oxinabox: oh it's not supposed to be better.
It is supposed to make you feel hate.

nickrobinson251: Ah cool, yeah it's a lot isn't it. Not my scene

PR #359, opened 2022-08-09
github expects `tags` field to contain a list

tpgillam:


tpgillam:
I am confused as to how this change could reduce coverage by so much! Is this a codecov issue, or is something else going on?

iamed2 (approved):

PR #362, opened 2022-08-19
Typo (push->push!) and type fix ("None"->:none)

contradict:


oxinabox:
Seems right, but do you have an example of when this causes issues?
As it is in the interactive part i don't really expect an automated test case nesc. but something I could manually run to veruify this was broken and now is fixed?

contradict:
Before:
```
julia> tt = Template(; user="contradict", dir=".", host="github.com", julia=v"1.8.0", interactive=true, plugins=[])
ERROR: UndefVarError: push not defined
```

After:
```
julia> tt = Template(; user="contradict", dir=".", host="github.com", julia=v"1.8.0", interactive=true, plugins=[])
Template keywords to customize:
[press: d=done, a=all, n=none]
 > [ ] authors
   [ ] none
```

oxinabox (approved):

PR #369, opened 2022-10-19
Migrate links to JuliaCI/PkgTemplates.jl

iamed2:
Most of these redirected but it's good to update them anyways. Looks like docs got migrated without having to be regenerated, but Codecov will probably be empty until we have a successful test run.


nickrobinson251 (approved):
thanks!

PR #372, opened 2022-10-26
gitlab-ci.yml file prefix

antonhinneck:
Hi,
PkgTemplates.jl suits my needs to generate new projects with CI pipeline perfectly.
However, I need to be able to make further modifications to gitlab-ci.yml files (set variables during package creation).
While a custom template could be created, addressing this directly in PkgTemplates.jl would be 
- extremly straightforward and 
- one would not have to maintain an additional template, beyond what is already in this package.

With this PR, one could exemplarily set the http_proxy using GitLabCI like so:

```
tpl = Template(;
    dir="./",
    host="host.com",
    user="antonhinneck",
    plugins=[Git(), 
    GitLabCI(;coverage=true,
    file_prefix="variables:\n\thttp_proxy: 10.0.0.1\n")]
)

tpl("Pkg.jl")
```

All tests pass. Would this be sensible for the scope of this package?

Best,
Anton

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,7 @@ version = "0.7.29"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
+DeepDiffs = "ab62b9b5-e342-54a8-a765-a90f495de1a6"
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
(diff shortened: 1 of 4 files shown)


gdalle:
I don't see a reason not to accept this. Can you add documentation and tests for this? The reference tests compare generated packages to a series of "fixtures", and they are only run for Julia 1.7.2. I suggest you modify the `WackyOptions` test suite accordingly

PR #379, opened 2023-01-18
Add `statuses: write` permission to GH token for docdeploy

fredrikekre:
This is needed to post the GitHub status with preview links, see [Documenter.jl documentation](https://documenter.juliadocs.org/dev/man/hosting/#GitHub-Actions).

### DIFF
--- a/templates/github/workflows/CI.yml
+++ b/templates/github/workflows/CI.yml
@@ -67,6 +67,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write
+      statuses: write
     steps:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
--- a/test/fixtures/DocumenterGitHubActions/.github/workflows/CI.yml
+++ b/test/fixtures/DocumenterGitHubActions/.github/workflows/CI.yml
@@ -39,6 +39,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write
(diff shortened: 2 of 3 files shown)


nickrobinson251 (approved):
Thanks! 

(apologies for the broken CI here)

nickrobinson251:
Oh but this failure is real: https://github.com/JuliaCI/PkgTemplates.jl/actions/runs/3949873044/jobs/6761646319

We just need to update the `test/fixtures/` files to what we expect

I think that means changing the two reference tests that have GHA CI and use Documenter i.e.
- https://github.com/JuliaCI/PkgTemplates.jl/blob/0de5d855e050d93169f8661a13b3a53a8cb2b283/test/fixtures/DocumenterGitHubActions/.github/workflows/CI.yml#L40
- https://github.com/JuliaCI/PkgTemplates.jl/blob/0de5d855e050d93169f8661a13b3a53a8cb2b283/test/fixtures/WackyOptions/.github/workflows/CI.yml#L40

Also, do we need to change this repo's own documenter setup in anyway (i.e. https://github.com/JuliaCI/PkgTemplates.jl/blob/0de5d855e050d93169f8661a13b3a53a8cb2b283/.github/workflows/CI.yml)?

fredrikekre:
> Also, do we need to change this repo's own documenter setup in anyway

Not necessarily. I believe by default the token has all permissions, so since it is not limited to `contents: write` it also can post statuses.

PR #384, opened 2023-02-08
Add CodeOwners Plugin

oxinabox:
This adds the CODEOWNERS file feature.
See [GitLab docs](https://docs.gitlab.com/ee/user/project/code_owners.html), and [GitHub docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners)


oxinabox:
1.0 is broken cos of https://github.com/JuliaCI/PkgTemplates.jl/issues/370

src/plugins/codeowners.jl
```diff
@@ -0,0 +1,39 @@
+"""
+    CodeOwners <: Plugin
+    CodeOwners(; fallback_owner)
```
iamed2, line 3: Does this constructor exist?

src/plugins/codeowners.jl
```diff
@@ -0,0 +1,39 @@
...
+
+A plugin which created GitLab/GitHub compatible CODEOWNERS files.
+owners should be a vector of patterns mapped to a vector of owner names.
+For example:
+`owners=["*"=>["@invenia"], "README.md"=>["@documentation","@oxinabox]]`
```
iamed2, line 8: Is this a vector (instead of a dict) because the file is semantically ordered?

test/fixtures/WackyOptions/CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+* @user
```
iamed2, line 1: I think this has semantic meaning that GitHub is parsing and so perhaps the test files should be kept ignored by this repository somehow. Unless these are explicitly reserved by GitHub for testing?

src/plugins/codeowners.jl
oxinabox: It does not. It is outdated from an earlier draft.

src/plugins/codeowners.jl
oxinabox: Correct.
Later has greater precedence than earlier

test/fixtures/WackyOptions/CODEOWNERS
oxinabox: It shouldn't act on this file, in this location.

> CODEOWNERS file location
>
> To use a CODEOWNERS file, create a new file called CODEOWNERS in the root, docs/, or .github/ directory of the repository, in the branch where you'd like to add the code owners.

From the GitHub docd

src/plugins/codeowners.jl
oxinabox:
```suggestion
    CodeOwners(; owners)
```

test/fixtures/WackyOptions/CODEOWNERS
morris25: If we wanted to be extra safe we could just point the references at files that don't exist.

morris25 (approved):

iamed2 (approved):

PR #377, opened 2023-01-16
Fix a typo in docs of `render_file`

singularitti:

PR #385, opened 2023-02-08
Only interactive tests pre-Julia 1.8

oxinabox:
Work around for https://github.com/JuliaCI/PkgTemplates.jl/issues/370

I think this is fine for now since our reference tests also require running in Julia 1.7.2 anyway.


oxinabox:
Merging this would let CI pass for #384 and #377

nickrobinson251 (approved):
thank you

PR #393, opened 2023-04-19
add dependabot plugin

ranocha:
I thought it may be nice to add Dependabot as default plugin. It's basically CompatHelper but for GitHub actions. See also https://github.com/JuliaCI/PkgTemplates.jl/pull/392 where I setup Dependabot for this repo.

This is my first real PR to this repo, so please tell me if there is anything else to do.


oxinabox:
Can we break this into two PRs?
One to just add this, (which can be this one)
and another to add it as a default plugin?

Adding this is a clear good, adding it as a default might be more debatable.
and i don't want the former to be blocked on the latter.

ranocha:
Sure, good point. I updated the PR accordingly.

src/plugins/dependabot.jl
```diff
@@ -0,0 +1,16 @@
+"""
+    Dependabot(; file="$(contractuser(default_file("github", "dependabot.yml")))")
+
+Setups Dependabot to create PRs whenever GitHub actions can be updated.
+This is very similar to [`CompatHelper`](@ref), which performs the same task
+for Julia package dependencies.
```
oxinabox, line 6: Need to mention here that this is (currently) only for the GithubActions [package-ecosystem](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file)


Since at somepoint someone might at Julia to Dependabot.
(which would be better than current CompatHelper as it doesn't time out after 60 days)

docs/src/user.md
```diff
@@ -55,6 +55,7 @@ AppVeyor
 CirrusCI
 DroneCI
 GitHubActions
+Dependabot
```
oxinabox, line 58: Is this Continuous Integration?
My not misc?

oxinabox (commented):
Other than the couple of comments on docs this seems fine

src/plugins/dependabot.jl
ranocha: Done in b88b1aa821c65c7eb8d0b17fa2c6c7a99584a1ba

docs/src/user.md
ranocha: I moved it in 70f01d0fd045c809dfffd003e3d33f1c0a396c4e

PR #396, opened 2023-04-24
make Dependabot a default plugin

ranocha:
As suggested in #393, I split the basic implementation of the Dependabot plugin from this PR setting it as a default plugin. My rationale for making it default is as follows.
- Some [default plugins](https://juliaci.github.io/PkgTemplates.jl/stable/user/#Default-Plugins-1) such as [CompatHelper](https://juliaci.github.io/PkgTemplates.jl/stable/user/#PkgTemplates.CompatHelper) and [TagBot](https://juliaci.github.io/PkgTemplates.jl/stable/user/#PkgTemplates.TagBot) use GitHub actions
- Updating dependencies seems to be considered important since [CompatHelper](https://juliaci.github.io/PkgTemplates.jl/stable/user/#PkgTemplates.CompatHelper) is a default plugin
- Thus, it seems reasonable to keep also GitHub actions dependencies up to date when using a default configuration


oxinabox:
I'm unclear how crucial this is.
Like it seems to me to create a lot of noise, requesting to update things that haven't changed much like simple actions such as `checkout`.
Though perhaps that noise goes away after you have been using it for a while.

This seems less important than CompatHelper, since the big issue for Compat is that it is easy to end up in a situation where two packages can't be used together due to requiring different versions of downstream packages.
This can't occur for github actions as we are not in julia packages defining new actions other people can use.
Once you have a set of actions that works, it will keep working forever, AIUI.
modulo it depending on external resources tat get taken down and/or changed.

I can imaging that last occuring for some of our julia CI actions, so maybe  that does make sense to have this turned on by default.

ranocha:
> Once you have a set of actions that works, it will keep working forever, AIUI.

That's not completely true as far as I know. For example, consider one of the latest CI runs in this repo:
https://github.com/JuliaCI/PkgTemplates.jl/actions/runs/4856935409

You can find lots of warnings

> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

As far as I understand, the old actions will not continue to work forever and we must update the versions of these actions to get something usable in the future. That's why I became aware of this issue and submitted PRs such as https://github.com/JuliaCI/PkgTemplates.jl/pull/392

gdalle:
I would be in favor of this. @oxinabox if you have no objections I'd like to get this PR up to date and merge it?

oxinabox:
I have no objections and if you are infavor of this then go ahead.

test/interactive.jl
```diff
@@ -133,7 +133,7 @@ end
                 CR, DOWN^5, CR, DONE,    # Customize user and plugins
                 USER, LF,                # Enter user
                 SELECT_DEFAULTS,         # Pre-select default plugins
-                UP^2, CR, UP^2, CR, DONE,# Disable TagBot and Readme
+                UP^3, CR, UP^2, CR, DONE,# Disable TagBot and Readme
```
gdalle, line 136: the number of moves up from the last default plugin necessary to reach TagBot changed

gdalle (approved):

PR #398, opened 2023-05-09
github & travis: dev package before building docs

musoke:
The Documenter.jl docs say to dev the package before building the docs.
- https://github.com/JuliaDocs/Documenter.jl/blob/233a7822d670b063e894f62aea00729bcb3a28dd/docs/src/man/hosting.md?plain=1#L201-L202

The gitlab-ci script already does this, so this commit increases consistency.

This is also discussed in https://github.com/JuliaDocs/Documenter.jl/issues/1565 and https://discourse.julialang.org/t/documenter-jl-is-not-documenting-any-changes/94781/.

I also added `/docs/Manifest.toml` to the default `.gitignore`; this may be more controversial. I am happy to revert that.

### DIFF
--- a/src/plugins/documenter.jl
+++ b/src/plugins/documenter.jl
@@ -111,7 +111,7 @@ defaultkw(::Type{<:Documenter}, ::Val{:index_md}) = default_file("docs", "src", 
 defaultkw(::Type{<:Documenter}, ::Val{:devbranch}) = nothing
 defaultkw(::Type{<:Documenter}, ::Val{:edit_link}) = :devbranch
 
-gitignore(::Documenter) = ["/docs/build/"]
+gitignore(::Documenter) = ["/docs/build/", "/docs/Manifest.toml"]
 priority(::Documenter, ::Function) = DEFAULT_PRIORITY - 1  # We need SrcDir to go first.
 
 badges(::Documenter) = Badge[]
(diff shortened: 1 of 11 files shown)


oxinabox:
This looks good to me.
You will need to update the reference files so the 1.7 tests pass.
Then we can merge it

musoke:
Oh, I missed that there are tests that only run on 1.7.2.  I have updated to include those now.


I also checked that the scripts work when used.  

github actions:
- built env as expected: https://github.com/musoke/TestDocEnv.jl/actions/runs/4952478706/jobs/8859002986#step:4:21 
- successfully deployed to https://musoke.github.io/TestDocEnv.jl/dev/
- docs successfully rebuilt when new commits were added: https://github.com/musoke/TestDocEnv.jl/actions/runs/4952537348/jobs/8859143936

travis:
- built env as expected: https://app.travis-ci.com/github/musoke/TestDocEnv.jl/jobs/602028274#L249-L253
- docs built successfully: https://app.travis-ci.com/github/musoke/TestDocEnv.jl/jobs/602028274#L249-L253
- didn't test deployment, but didn't change anything there

In both cases julia 1.0 failed, but that's because I ran PkgTemplates with `Tests(project=true)`.


There is one test that failed for me locally, but it fails on master too.  (Interactive mode -> Simulated inputs -> Quotes)

oxinabox:
> Oh, I missed that there are tests that only run on 1.7.2.

Yes, its a problem. We need them because
1. testing exact output with reference files ends up including julia versions
2. Julia 1.8 changed how tests interact with IO making it harder to test interactive workflow

It would be great if all tests are run on all version.
or at very least if those tests in  CI were labelled differently.

PR #399, opened 2023-05-12
make GitHubActions a default plugin

musoke:
Fixes https://github.com/JuliaCI/PkgTemplates.jl/issues/357

### DIFF
--- a/docs/src/user.md
+++ b/docs/src/user.md
@@ -41,6 +41,7 @@ Tests
 Readme
 License
 Git
+GitHubActions
 CompatHelper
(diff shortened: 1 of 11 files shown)


musoke:
It certainly wasn't meant to be breaking, but it does seem that you have identified a breakage.  I don't think the tests check any cases without these variables set.

oxinabox:
Possibly we should revert this?

Maybe this is a good excuse to go to v1.0.0

musoke:
I will defer to you on that.

PR #409, opened 2023-06-13
Make PkgEval badge link to this package's evaluation

gerlero:
As a suggestion, make the PkgEval badge link directly to the evaluation page for this package, instead of linking to the full report of all packages. This is better IMO and is also how the official repo recommends it at https://github.com/JuliaCI/NanosoldierReports#badges.


oxinabox:
thanks!

Issue #324, opened 2021-10-27
Add support for JuliaFormatter?


devmotion:
More and more packages seem to use JuliaFormatter to enforce a specific code style. Maybe support for JuliaFormatter could be added to PkgTemplates? This could mean e.g. to support creating a `.JuliaFormatter.toml` file with some default style (e.g., `style = blue`, in particular if the Blue style badge is used) and a [Github action](https://github.com/JuliaDiff/ChainRulesCore.jl/blob/main/.github/workflows/format.yml) that checks the format and provides suggestions in PRs (only if not made from a fork though: https://github.com/JuliaDiff/ChainRulesCore.jl/issues/489).

chriselrod:
I'd like to second this.
I've configured my editor to format on save, and for some reason the default formatter settings it uses are really bad (i.e., 8 spaces per tab), so I have to create a `.JuliaFormatter.toml` file in every repo where I want to save a file.

Of course, I could turn off the auto formatter, or maybe I could get it to only format if it finds a `.JuliaFormatter.toml`...
But for now, I think it'd be nice if `.JuliaFormatter.toml`s were more widespread.

nickrobinson251:
A PR would be welcomed, if anyone can be tempted to give it a go :)

Issue #328, opened 2021-11-08
Add JuliaFormatter toml generator (maybe consistent with "code style")


jlperla:
Now that the VS Code has moved to JuliaFormatter.jl, I would be great for a template plugin to generate files to support it (including your beloved `style = "blue"`!)  

It seems the architecture is to remove all formatting in the vscode extension itself, and do all formatting configuration in a 
`.JuliaFormatter.toml` file (cc @pfitzseb).  Which I fully agree with!  That means it is especially helpful to generate these files for new users.

The minimum configuration of a plugin to the PkgTemplates is to generate a file with `style = "blue"`, `="default"` or `="yas"`.   Personally that is enough for my purposes since I think that users can then modify it directly since it would be created in the correct place.  But to get fancy you could add in https://github.com/domluna/JuliaFormatter.jl/#formatting-options  as well.

BTW, my hope is to add this into the https://julia.quantecon.org/software_engineering/testing.html#project-setup  docs so that introductory users get in the habit of always using code formatting/code styles.

nickrobinson251:
Sure, I think a `JuliaFormatter <: Plugin` would be a great addition!

gdalle:
I accidentally duplicated this in #324

gdalle:
> BTW, my hope is to add this into the https://julia.quantecon.org/software_engineering/testing.html#project-setup docs so that introductory users get in the habit of always using code formatting/code styles.

I'm currently on a rampage of PkgTemplates PR to teach people good coding practices. See #415 about my dreams for the `runtests.jl` file. If you have other similar ideas, I'll take 'em!

nickrobinson251:
closed by https://github.com/JuliaCI/PkgTemplates.jl/pull/412

PR #412, opened 2023-07-28
Add JuliaFormatter plugin

gdalle:
Add a JuliaFormatter plugin which takes a `style` keyword argument and constructs a `.JuliaFormatter.toml` file using only the style choice (it's up to the user to customize it further).

This sets the style of automatic formatting for [JuliaFormatter.jl](https://github.com/domluna/JuliaFormatter.jl), and thus also for the VSCode extension.

Solves #324


src/plugins/juliaformatter.jl
```diff
@@ -0,0 +1,26 @@
...
+This file is used by [JuliaFormatter.jl](https://github.com/domluna/JuliaFormatter.jl) and the Julia VSCode extension to configure automatic formatting.
+It can be further customized by the user beyond the style name, see the [JuliaFormatter.jl docs](https://domluna.github.io/JuliaFormatter.jl/stable/).
+
+## Keyword Arguments
+- `file::String`: Template file for `.JuliaFormatter.toml`.
+- `style::String`: Compatible style name (either "sciml", "blue" or "yas")
+"""
+@plugin struct JuliaFormatter <: FilePlugin
+    file::String = default_file(".JuliaFormatter.toml")
+    style::String = "sciml"
```
nickrobinson251, line 18: i don't think this should default to `sciml` style. Does JuliaFormatter have a default style? if so, i think it should default to that. If not, i think we should require users pass `style`

nickrobinson251 (commented):
Apart from changing the default, i think this looks great! please could you also increment the version number in Project.toml? thanks!

src/plugins/juliaformatter.jl
gdalle: Can I do a style attribute that is either `nothing` or a string? Will that work in interactive mode?

nickrobinson251: yes, i think that's fine -- `view` just needs to know what to do with `nothing` i think

gdalle: and how do I force the user to actively choose an option? the `@plugin` macro forces me to pick a default value for the attribute

nickrobinson251: i think maybe don't use `@plugin` and instead manually implement the necessary methods

See the docs on "Implementing `@plugin` manual" https://juliaci.github.io/PkgTemplates.jl/dev/developer/#PkgTemplates.@plugin

I think the License and Documenter plugins are examples of this

gdalle: I went for an empty string default and added a comment at the top of the created TOML file indicating how to fill it

gdalle: I think we're only missing some customization of the radio menu

nickrobinson251: > we're only missing some customization of the radio menu

What would you like it to look like? I think overloading `prompt` (and not using `@plugin`) gives you full customization, if that's what you'd like :)

src/plugins/juliaformatter.jl
```diff
@@ -0,0 +1,33 @@
...
+destination(::JuliaFormatter) = ".JuliaFormatter.toml"
+
+function view(p::JuliaFormatter, t::Template, pkg::AbstractString)
+    d = Dict{String,String}()
+    if p.style == ""
+        d["STYLE"] = ""
+    elseif p.style in ("blue", "sciml", "yas")
+        d["STYLE"] = """style = \"$(p.style)\""""
+    else
+        throw(ArgumentError("Formatting style not recognized"))
```
nickrobinson251, line 30: i suppose an issue with this is that it errors are template-usage time, not template-construction time... i wonder if it'd be more user-friendly to require one of the known styles at construction time -- what do you think?

gdalle: Agreed, but I wasn't sure how to define a constructor within the `@plugin` macro

gdalle: In fact, I don't think I can

gdalle: ooooh I can use `validate`

src/plugins/juliaformatter.jl
gdalle: done :)

gdalle:
This is ready!

src/plugins/juliaformatter.jl
```diff
@@ -0,0 +1,45 @@
...
+        throw(ArgumentError("""JuliaFormatter style must be either "nostyle", "blue", "sciml" or "yas"."""))
+    end
+end
+
+source(p::JuliaFormatter) = p.file
+destination(::JuliaFormatter) = ".JuliaFormatter.toml"
+
+function view(p::JuliaFormatter, t::Template, pkg::AbstractString)
+    d = Dict{String,String}()
+    if p.style == "nostyle"
```
nickrobinson251, line 31: why not use `nothing` for this?

gdalle:
Also solves #328

For now the BlueStyle plugin is separate from the JuliaFormatter one, perhaps we should deprecate it and add a generic code style badge based on the JuliaFormatter style?

nickrobinson251 (approved):

src/plugins/juliaformatter.jl
nickrobinson251: how about just the empty string `""`?

nickrobinson251:
thanks, again!

gdalle:
You're welcome! Good for merging?

src/PkgTemplates.jl
```diff
@@ -33,6 +33,7 @@ export
     Git,
     GitHubActions,
     GitLabCI,
+    JuliaFormatter,
```
nickrobinson251, line 36: one problem with this that i just thought of, is that if JuliaFormatter.jl is loaded (e.g. in `startup.jl` ), then we'll see
```julia
julia> using PkgTemplates
WARNING: using PkgTemplates.JuliaFormatter in module Main conflicts with an existing identifier.
```

...do people put JuliaFormatter in their startup?

Do you think this is an issue?

nickrobinson251: I'm not a regular JuliaFormatter user so i've no idea if this name conflict is going to be a common issue for users

but there's only one formatter for Julia... so i wonder if we should just name this plugin `Formatter` (or `JuliaFormat` or some such) to avoid the name conflict?

(I guess we don't have this issue with `Documenter` as it's usually only loaded in a separate `docs/` env)

gdalle: I think renaming to Formatter is a good idea, I have JuliaFormatter in my startup

gdalle: done in the next commit

gdalle:
Not sure how to deprecate the BlueStyle badge?

nickrobinson251:
if you bump the version we can make a release straight away :)

gdalle:
Done!

gdalle:
Only bumped the patch version because I believe the change is nonbreaking

> In particular, a package may set version = "0.2.4" when it has feature additions compared to 0.2.3 as long as it remains backward compatible with 0.2.0. See also [The version field](https://pkgdocs.julialang.org/v1/toml-files/#The-version-field).

https://pkgdocs.julialang.org/v1/compatibility/#compat-pre-1.0

PR #413, opened 2023-07-28
Make testsets verbose

gdalle:
Really just a find+replace of `@testset` with `@testset verbose = true` for more granularity in the test results.

I took care not to modify the `runtests.jl` of the templates, this change is restricted to the actual tests of PkgTemplates.jl itself


gdalle:
Oops it's not supported by 1.0. Nevermind I guess

PR #414, opened 2023-07-28
Explain reference Julia version for tests

gdalle:
Add a paragraph to the developer guide recommending local tests against Julia 1.7.2, otherwise the reference tests will not run and they are pretty essential.

I made the mistake myself and only discovered the test failure in the PR CI


gdalle:
Not sure why one test was cancelled, it's a pure docs update

nickrobinson251 (approved):

Issue #411, opened 2023-07-24
Add Aqua to tests


gdalle:
[Aqua.jl](https://github.com/JuliaTesting/Aqua.jl) is cementing itself as an important tool for testing the quality of a code base.
It would be cool to have an option that generates `test/runtests.jl` with Aqua built-in, like so:

```julia
using Aqua
using MyPackage

@testset verbose = true "MyPackage.jl" begin
    @testset verbose = true "Aqua.jl" begin
        Aqua.test_all(MyPackage)
    end
    # actual tests
end

gdalle:
@nickrobinson251 I tried my hand at an easier plugin with #412, but for this one I'd actually need to modify an existing file created by the test plugin instead of creating it from scratch. Any good example I should look at?

gdalle:
Crap I also need to modify a whole bunch of TOMLs because it requires Aqua as a test dependency

gdalle:
Gonna try to do this by adding an `aqua` option to the `Tests` plugin

PR #415, opened 2023-07-31
Add Aqua to the Tests plugin

gdalle:
Solves #411 by adding a boolean field named `aqua` to the `Tests` plugin.
When it is set to `true`:

- Aqua is added as a test-specific dependency, either in `Project.toml` or `test/Project.toml`

- The file `test/runtests.jl` is modified to look like this:
```julia
using WackyOptions
using Test, Aqua

@testset "WackyOptions.jl" begin
    # Write your tests here.
    @testset verbose = true "Code quality (Aqua.jl)" begin
        Aqua.test_all(WackyOptions)
    end
end
```

If this is approved, I would like to do the same for other standard test tools (JET and JuliaFormatter, once #412 is merged).
I can either add them to this PR or put them in another one.

A few doubts I have:

- The templating behavior in `templates/test/runtests.jl` is not very elegant. Basically I put the placeholders at the end to avoid creating useless empty lines.
- As you can see from the `test/fixtures/WackyOptions/test/Manifest.toml` diff, Aqua is the first dependency that is not from the standard library. So whenever Aqua releases a new stable release, the reference tests will break because `Pkg.add("Aqua")` will pick the latest one. No sure how to future-proof that.
- There is nowhere I can test this with `project=false`, so I'm not entirely sure I got the mechanics of test-specific dependencies in the main `Project.toml` completely right.


src/plugins/tests.jl
```diff
@@ -19,44 +27,83 @@ Sets up testing for packages.
 @plugin struct Tests <: FilePlugin
     file::String = default_file("test", "runtests.jl")
     project::Bool = false
+    aqua::Bool = false
```
hannahilea, line 30: Thanks for adding this! 

In many of my projects, I set aqua but with a different set of default kwargs (often `ambiguities=false`). Perhaps we could add an additional field to the `Tests` struct that allows a user to set those kwargs up front? 
e.g., 
```suggestion
    aqua::Bool = false
    aqua_kwarg_str::String = ""
```
and then (below)
```
kwargs = isempty(p.aqua_kwarg_str) ? "" : "; $(p.aqua_kwarg_str)"
Aqua.test_all($pkg$kwargs)
```

(or a sleeker version that passes the params through as a NamedTuple! But this easy version would be sufficient for my needs)

gdalle: done in the next commit :)

gdalle:
Beware that I removed Manifest comparison to ensure that non standard dependencies like Aqua can release new versions without breaking the PlgTemplates test pipeline. Is there a better way?

src/plugins/tests.jl
hannahilea: thanks!

nickrobinson251:
hey, thanks for this! i don't have time to review right now. i wonder if @hannahilea would be happy to review this one to completion?

hannahilea:
sorry, i've been on vacation---i'll be back and happy to review this on wednesday if no one gets to it first!

gdalle:
No one has to apologize for being on vacation! Enjoy the rest of it, there is no rush

src/plugins/tests.jl
```diff
@@ -19,44 +29,106 @@ Sets up testing for packages.
...
+        add_test_dependency(p, pkg_dir)
+    end
 end
 
 # Create a new test project.
-function make_test_project(pkg_dir::AbstractString)
+function make_test_project(p::Tests, pkg_dir::AbstractString)
     with_project(() -> Pkg.add(TEST_DEP), joinpath(pkg_dir, "test"))
+    if p.aqua
+        with_project(() -> Pkg.add(AQUA_DEP), joinpath(pkg_dir, "test"))
```
hannahilea, line 102:
```suggestion
        with_project(() -> Pkg.add(; AQUA_DEP.name, AQUA_DEP.uuid, version=p.aqua_version), joinpath(pkg_dir, "test"))
```

src/plugins/tests.jl
```diff
@@ -19,44 +29,106 @@ Sets up testing for packages.
...
 
 # Add Test as a test-only dependency.
-function add_test_dependency(pkg_dir::AbstractString)
+function add_test_dependency(p::Tests, pkg_dir::AbstractString)
     # Add the dependency manually since there's no programmatic way to add to [extras].
     path = joinpath(pkg_dir, "Project.toml")
     toml = TOML.parsefile(path)
+    if p.aqua
+        get!(toml, "extras", Dict())["Aqua"] = AQUA_UUID
+    end
```
hannahilea, line 113:
```suggestion
    if !isnothing(p.aqua)
        get!(toml, "extras", Dict())["Aqua"] = AQUA_UUID
    end
    if p.aqua isa VersionNumber
        get!(toml, "compat", Dict())["Aqua"] = "=$(p.aqua)"
    end
```

src/plugins/tests.jl
```diff
@@ -19,44 +29,106 @@ Sets up testing for packages.
...
     get!(toml, "extras", Dict())["Test"] = TEST_UUID
-    get!(toml, "targets", Dict())["test"] = ["Test"]
+    get!(toml, "targets", Dict())["test"] = p.aqua ? ["Aqua", "Test"] : ["Test"]
```
hannahilea, line 115:
```suggestion
    get!(toml, "targets", Dict())["test"] = isnothing(p.aqua) ? ["Test"] : ["Aqua", "Test"]
```

test/reference.jl
```diff
@@ -74,9 +74,12 @@ function test_all(pkg::AbstractString; kwargs...)
         pkg_dir = joinpath(t.dir, pkg)
         PT.hasplugin(t, Documenter) && pin_documenter(joinpath(pkg_dir, "docs"))
         foreach(readlines(`git -C $pkg_dir ls-files`)) do f
-            reference = joinpath(@__DIR__, "fixtures", pkg, f)
-            comparison = joinpath(pkg_dir, f)
-            test_reference(reference, comparison)
+            if !contains(f, "Manifest.toml")
+                # Don't check Manifest: versions of packages like Aqua may vary
```
hannahilea, line 78:
```suggestion
```
...and can also remove corresponding `end`. If you choose to make `p.aqua` include the aqua version used (or otherwise pass version in), we can set the version in the test and then continue to test the manifest.

test/reference.jl
```diff
@@ -153,7 +156,11 @@ end
                 dispatch=true,
                 dispatch_delay=20,
             ),
-            Tests(; project=true),
+            Tests(;
+                project=true,
+                aqua=true,
```
hannahilea, line 161:
```suggestion
                aqua_version=v"0.6.5",
```
to fix it so that the manifest doesn't change!

src/plugins/tests.jl
```diff
@@ -19,44 +29,106 @@ Sets up testing for packages.
...
+        else
+            kwargs_str = "; " * strip(string(p.aqua_kwargs), ['(', ')'])
+        end
+        d["AQUA_IMPORT"] = ", Aqua"
+        d["AQUA_TESTSET"] = """
+
+            @testset verbose = true "Code quality (Aqua.jl)" begin
+                Aqua.test_all($pkg$kwargs_str)
+            end
+        """[1:end-1]
```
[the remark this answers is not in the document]
hannahilea: (updated formatting for moving the test above the space for additional tests---see test for example of updated formatting)

src/plugins/tests.jl
```diff
@@ -19,44 +29,106 @@ Sets up testing for packages.
...
+        :deps_compat,
+        :project_toml_formatting,
     )
+    for (key, val) in pairs(p.aqua_kwargs)
+        if !(val isa Bool)
+            throw(ArgumentError("Aqua keyword arguments must have boolean values"))
+        elseif !(key in aqua_kwargs_names)
+            throw(ArgumentError("Aqua keyword arguments must belong to $aqua_kwargnames"))
+        end
+    end
```
hannahilea, line 83:
```suggestion
    end
    _validate_aqua_version(p.aqua_version)
```

and then outside this function:
```
_validate_aqua_version(::Nothing) = nothing
_validate_aqua_version(::VersionNumber) = nothing
function _validate_aqua_version(v::AbstractString)
    v == "*" || throw(ArgumentError("Aqua version must be `VersionNumber`, `"*"` for the latest version of Aqua.jl, or `nothing` to skip use of Aqua."
    return nothing
end

```

src/plugins/tests.jl
```diff
@@ -19,44 +29,106 @@ Sets up testing for packages.
...
+    else
+        add_test_dependency(p, pkg_dir)
+    end
 end
 
 # Create a new test project.
-function make_test_project(pkg_dir::AbstractString)
+function make_test_project(p::Tests, pkg_dir::AbstractString)
     with_project(() -> Pkg.add(TEST_DEP), joinpath(pkg_dir, "test"))
+    if p.aqua
```
hannahilea, line 101:
```suggestion
    if !isnothing(p.aqua_version)
```

src/plugins/tests.jl
```diff
@@ -19,44 +29,106 @@ Sets up testing for packages.
...
+    end
 end
 
 # Add Test as a test-only dependency.
-function add_test_dependency(pkg_dir::AbstractString)
+function add_test_dependency(p::Tests, pkg_dir::AbstractString)
     # Add the dependency manually since there's no programmatic way to add to [extras].
     path = joinpath(pkg_dir, "Project.toml")
     toml = TOML.parsefile(path)
+    if p.aqua
```
hannahilea, line 111:
```suggestion
    if !isnothing(p.aqua_version)
```

src/plugins/tests.jl
```diff
@@ -19,44 +29,106 @@ Sets up testing for packages.
...
     write_project(path, toml)
 
     # Generate the manifest by updating the project.
     touch(joinpath(pkg_dir, "Manifest.toml"))  # File must exist to be modified by Pkg.
     with_project(Pkg.update, pkg_dir)
 end
+
+function badges(p::Tests)
+    bs = Badge[]
+    if p.aqua
```
hannahilea, line 125:
```suggestion
    if !isnothing(p.aqua_version)
```

test/show.jl
```diff
@@ -88,6 +88,8 @@ end
                 Tests:
                   file: "$(joinpath(TEMPLATES_DIR, "test", "runtests.jl"))"
                   project: false
+                  aqua: false
```
hannahilea, line 91:
```suggestion
                  aqua_version: nothing
```

hannahilea (changes requested):
This looks great!

I added a couple style preferences for ordering in the resultant runtests.jl file---feel free to ignore those.

> Beware that I removed Manifest comparison to ensure that non standard dependencies like Aqua can release new versions without breaking the PlgTemplates test pipeline. Is there a better way?

What if instead of the `aqua` kwarg, we instead use a `aqua_version` kwarg, with options `nothing` (don't use Aqua), `"*"` (use latest version of Aqua), or `v"0.6.2"` (set a compat bound equal to the provided version number)? I added suggestions to this affect---if you'd prefer not to do this, another option could be to add an additional kwarg that sets the version, so that (again) the version could be fixed for the manifest testing.

I have some ideas for testing the otherwise-untested code paths, but those will have to wait for tomorrow.

src/plugins/tests.jl
```diff
@@ -19,44 +29,106 @@ Sets up testing for packages.
...
+
+function view(p::Tests, ::Template, pkg::AbstractString)
+    d = Dict("PKG" => pkg)
+    if p.aqua
+        if isempty(p.aqua_kwargs)
+            kwargs_str = ""
+        else
+            kwargs_str = "; " * strip(string(p.aqua_kwargs), ['(', ')'])
+        end
+        d["AQUA_IMPORT"] = ", Aqua"
```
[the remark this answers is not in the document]
hannahilea: (adds imports on separate lines)

src/plugins/tests.jl
```diff
@@ -19,44 +29,106 @@ Sets up testing for packages.
 @plugin struct Tests <: FilePlugin
     file::String = default_file("test", "runtests.jl")
     project::Bool = false
+    aqua::Bool = false
```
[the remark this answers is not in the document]
gdalle: Interesting workaround! But I'm afraid that if we're too strict, it might freeze Aqua in the version specified by the user when PkgTemplates was called, instead of letting it update to the latest available release.
I'll read the rest of your comments and come back to this one

src/plugins/tests.jl
```diff
@@ -19,44 +29,106 @@ Sets up testing for packages.
...
+    if p.project
+        make_test_project(p, pkg_dir)
+    else
+        add_test_dependency(p, pkg_dir)
+    end
 end
 
 # Create a new test project.
-function make_test_project(pkg_dir::AbstractString)
+function make_test_project(p::Tests, pkg_dir::AbstractString)
```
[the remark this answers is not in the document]
gdalle: agreed

src/plugins/tests.jl
```diff
@@ -19,44 +29,106 @@ Sets up testing for packages.
...
     with_project(() -> Pkg.add(TEST_DEP), joinpath(pkg_dir, "test"))
+    if p.aqua
+        with_project(() -> Pkg.add(AQUA_DEP), joinpath(pkg_dir, "test"))
+    end
 end
 
 # Add Test as a test-only dependency.
-function add_test_dependency(pkg_dir::AbstractString)
+function add_test_dependency(p::Tests, pkg_dir::AbstractString)
```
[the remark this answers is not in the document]
gdalle: yep

test/fixtures/WackyOptions/test/runtests.jl
```diff
@@ -1,6 +1,9 @@
 using WackyOptions
-using Test
+using Test, Aqua
 
 @testset "WackyOptions.jl" begin
     # Write your tests here.
+    @testset verbose = true "Code quality (Aqua.jl)" begin
```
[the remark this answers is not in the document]
gdalle: okay

src/plugins/tests.jl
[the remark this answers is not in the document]
gdalle: okay this works with the `"*"` option

src/plugins/tests.jl
gdalle: good one

gdalle:
@hannahilea thanks for the feedback, I took your suggestions into account... but it still wasn't enough.

With the new `aqua_version` field, we control the specific version of Aqua that is installed into our test `Manifest.toml`, but we do not control the specific versions of _Aqua's dependencies_. This is why the reference tests fail due to a new version of the [Compat](https://github.com/JuliaLang/Compat.jl) package, on which Aqua depends. Not sure how to deal with that.

hannahilea:
> With the new aqua_version field, we control the specific version of Aqua that is installed into our test Manifest.toml, but we do not control the specific versions of Aqua's dependencies. This is why the reference tests fail due to a new version of the [Compat](https://github.com/JuliaLang/Compat.jl) package, on which Aqua depends. Not sure how to deal with that.

oof, brutal.  😂 okay, let me think about it a bit more...

hannahilea:
Okay. So I think the reference tests are important enough that we shouldn't remove them, so what if we:
1. don't test aqua in the "Wacky" test at all (set it to `nothing`---or `false`, if you'd prefer to revert back to the version-less Aqua)
2. added separate tests for the added aqua functionality, and specifically don't test the manifest for those?

gdalle:
The thing is, I'm planning to add other things to the test template, like JuliaFormatter or JET. So we'll face that problem again soon. Duplicated because we also need to test the "extras" method for test dependencies. 
Im not sure why it is so crucial to include manifests in the reference tests? Why aren't projects enough?

gdalle:
By the way, it's probably also the Manifest testing that forces us to only run reference tests on a fixed Julia version (1.7.2), which is kind of bizarre.

hannahilea:
> Im not sure why it is so crucial to include manifests in the reference tests? Why aren't projects enough?

Yeah, that's a good point, and I think I agree that removing manifests is probably the right call. Maybe projects _are_ enough! I haven't contributed enough to this package to make that decision; maybe @nickrobinson251 has an opinion (or could point us to someone who does)?

gdalle:
If we do get rid of them manifests (probably best done in a separate PR), I think I'll roll back your changes regarding the aqua version to simplify user experience

hannahilea:
yeah, i agree---no need to specify version there if we don't actually need it.

hannahilea:
(wish i'd thought about the full set of dependencies in the manifest before sending us down this path---ah well, next time i'll know better!)

gdalle:
@hannahilea I think we're good to ship this

hannahilea (approved):
Agreed!

PR #419, opened 2023-08-07
Support duplicate plugins

### LINKED ISSUE #266: Allow multiple plugins of the same type (e.g. `GithubActions` CI plugins)
Currently, if two plugins have the same type, the first plugin is used, see e.g. https://github.com/invenia/PkgTemplates.jl/pull/265#issuecomment-772536740 Why is that?

My use-case for multipl plugins of the same type is multiple `GitHubAction` CI plugins, e.g. for when if you want a different CI job for running on the Julia Nightly versions, or something like that. (Noticed at #265)

I think this is implemented at
https://github.com/invenia/PkgTemplates.jl/blob/238bfc11c037f341eff4889e0424a66d1edf7404/src/template.jl#L100
but i don't know why we do that.

I think it might be so that we can use a pattern like
```julia
for T in PluginTypes
   p = getplugin(tpl, T)
   ...
end
```
but i think that only appears one ([in the code for `view(p::Readme, ...`](https://github.com/invenia/PkgTemplates.jl/blob/238bfc11c037f341eff4889e0424a66d1edf7404/src/plugins/readme.jl#L30-L39)), and we could easily get the desired behaviour (sorting by the type `T` somehow) by writng the code slightly differently.

Is there any other reason to disallow multiple plugins of the same type?

hannahilea:
Closes #266.

Re-adds the changes started by @nickrobinson251 in #269, with the addition of tests.

Question: Previously, duplicated plugins would be silently removed. Now, they're silently allowed---even in cases where they could potentially lead to undefined behavior (in cases where only one plugin of a certain type is implicitly expected). Not sure what this means for semver: since it's a change in behavior, perhaps it should be a breaking change? But since it was never an explicitly stated behavior, perhaps it's just unbreaking? ...or maybe we should support a flag for falling back on the old behavior? Or throw a warning if duplicate plugins exist? I have no strong opinion here.

- [x] TODO: update version once decision on semver status is decided.


nickrobinson251:
i'm inclined towards saying this isn't breaking; i think duplicate plugins was undefined behaviour before (we never stated which would "win"), so i don't think anyone could safely be relying on passing duplicate plugins.

nickrobinson251 (approved):
LGTM

Thanks for picking this up!

hannahilea:
great! i bumped the patch version so that once this is merged we can tag a release (since it looks like this packages makes frequent releases on a per-PR basis anyway). 

also i was expecting to have to do a significant testing lift or hunt down edge cases, but it turns out your initial implementation was ready to go, and really did just need the few lines of tests.... so thank you!

hannahilea:
...not sure why codecov dropped here, but it seems unrelated to this PR. I think this is ready to merge!

nickrobinson251:
yeah, that's fine -- please could you rebase on master? (another PR went in first, so the version will need bumping one more)

hannahilea:
done!

PR #421, opened 2023-08-28
Do not truncate ".jl" from generated package directory

### LINKED ISSUE #353: Hope Name generated dir MyPkg.jl rather than MyPkg
using PkgTemplates
Template(interactive=true)("MyPkg.jl")

then generate the dir is MyPkg, why not MyPkg.jl ??

hannahilea:
Fixes #353.

Probably non-breaking, as I don't see it explicitly documented anywhere that a ".jl" suffix _would_ be removed when generating a package.

Updated approach respects user input: `Template("MyPkg.jl")` will create directory "MyPkg.jl", while `Template("MyPkg")` will generate directory "MyPkg". Contents of these two packages will be otherwise identical.

### DIFF
--- a/src/plugin.jl
+++ b/src/plugin.jl
@@ -267,7 +267,7 @@ prehook(::Plugin, ::Template, ::AbstractString) = nothing
 Stage 2 of the package generation pipeline (the "main" stage, in general).
 At this point, the [`prehook`](@ref)s have run, but not the [`posthook`](@ref)s.
 
-`pkg_dir` is the directory in which the package is being generated; [`pkg_name_from_pkg_dir`](@ref)`
+`pkg_dir` is the directory in which the package is being generated; [`pkg_name`](@ref)`
 will return the package name.
 
 !!! note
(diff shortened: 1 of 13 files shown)


src/plugin.jl
```diff
@@ -346,6 +346,18 @@ you should implement this function and return `true`.
...
 needs_username(::Plugin) = false
 
+"""
+    pkg_name_from_pkg_dir(pkg_dir::AbstractString)
+
+Return package name of package at `pkg_dir`, i.e., `basename(pkg_dir)` excluding any
+`.jl` suffix, if present. For example, `foo/bar/Whee.jl` and `foo/bar/Whee` both
+return `Whee`.
+"""
+function pkg_name_from_pkg_dir(pkg_dir::AbstractString)
```
nickrobinson251, line 356: maybe just name this `pkg_name`, since it's called like `pkg_name(pkg_dir)`? https://github.com/invenia/BlueStyle#function-naming

```suggestion
function pkg_name(pkg_dir::AbstractString)
```

src/plugin.jl
```diff
@@ -346,6 +346,18 @@ you should implement this function and return `true`.
...
+    pkg = basename(pkg_dir)
+    return endswith(pkg, ".jl") ? pkg[1:end-3] : pkg
```
nickrobinson251, line 358: we could use `chopsuffix` (but idk what Julia version that was introduced in, so maybe we can't)

```suggestion
    return chopsuffix(pkg, ".jl")
```

nickrobinson251 (approved):
thanks!

src/plugin.jl
hannahilea: TIL! Looks like it shows up in 1.8...

src/plugin.jl
hannahilea: done!

PR #422, opened 2023-08-31
Remove Manifests from reference tests

gdalle:
Following the discussion in #415, the reference test suite needs weakening.
As soon as PkgTemplates.jl installs dependencies that fall outside the standard library, it is impossible (or very cumbersome) to control that a specific version of _every subdependency_ is selected. As a result, I only include `Project.toml` files in the reference tests, and no longer `Manifest.toml` files. In fact, the latter are dropped from the repo altogether thanks to `.gitignore`.

Ping @hannahilea, this is an easy review


nickrobinson251:
do you know why we were ever including these?

gdalle:
> do you know why we were ever including these?

I wasn't there when the ancient code was written, but I can't think of a compelling reason

nickrobinson251:
could you please take a quick look in the history to check there wasn't a good reason

otherwise, i've no objection to this

gdalle:
For one thing there is https://github.com/JuliaCI/PkgTemplates.jl/pull/143: it's useful to check that the parent package is indeed `dev`-ed in the docs environment, and this cannot be seen from the Project alone

nickrobinson251:
nice find! i think that's maybe an important test -- can we keep that test somehow? e.g. keep the `docs/Manifest.toml` (perhaps with a more limited test that just checks for the presence of the package we expect to be dev'ed)?

nickrobinson251 (approved):
thanks!

hannahilea:
Nice!

Might be worth filing issue so that adding back coverage of https://github.com/JuliaCI/PkgTemplates.jl/pull/422#issuecomment-1701060143 isn't lost

nickrobinson251:
don't we still have that coverage, because we kept the `docs/Manifest.toml`? Or are we filtering them out too? As far as i understand, we should be able to guarantee the docs/Manifest.toml versions (for a fixed Documenter version), no?

hannahilea:
You're right---we're only filtering out `joinpath("test", "Manifest.toml"))`, not  all "Manifest.toml". Good!

gdalle:
I think it is related to https://github.com/JuliaCI/PkgTemplates.jl/issues/423

PR #426, opened 2023-09-02
Don't create empty Manifest.toml (again)

cossio:
Follow-up to https://github.com/JuliaCI/PkgTemplates.jl/pull/425. The `Git` plugin was also `touch`ing an empty Manifest.toml which also triggered the warning. This fixes that.


nickrobinson251:
Yes that failure is real -- please see https://juliaci.github.io/PkgTemplates.jl/stable/developer/#Testing-1 and update the reference tests as needed (which you can do by running the reference tests locally with the required Julia version)

nickrobinson251 (approved):

PR #431, opened 2023-10-02
update to actions/checkout@v4

visr:
A simple update of [`checkout`](https://github.com/actions/checkout) from v3 to v4.

As long as Dependabot is not used by default (#396) these often won't be updated after creation, so good to keep them up to date here.


gdalle:
Duplicate of #427 ?

gdalle (approved):

PR #441, opened 2023-12-04
Use CI caching for documentation job and add newly required perms

IanButterworth:
See new perm requirements in https://github.com/julia-actions/cache/releases/tag/v1.4.0

Note that they're not mandatory, just that tidying up happens more efficiently if they are in place, and lowers the chance of a useful cache being auto-deleted by Github


oxinabox (approved):

gdalle:
This PR failed reference tests because the fixtures were not updated. I'll fix it

Issue #434, opened 2023-10-12
Quote test is failing locally and ignored on CI


gdalle:
In the interactive test suite, one test always fails in `Simulated inputs / Quotes`.
The `user` and `dir` keyword arguments to `Template` are given with escaped quotes:

```julia
@testset "Quotes" begin
    print(
        stdin.buffer,
        CR, DOWN^2, CR, DONE,  # Customize user and dir
        "\"me\"", LF,          # Enter user with quotes
        "\"~\"", LF,           # Enter home dir with quotes
    )
    t1 = Template(; interactive=true)  # debug addition
    t2 = Template(; user="me", dir="~")  # debug addition
    @warn "Quotes" t1.user t1.dir t2.user t2.dir  # debug addition
    result = Template(; interactive=true) == Template(; user="me", dir="~")
    if get(ENV, "CI", "false") == "true"
        @test_broken result
    else
        @test result
    end
    [...]
end
```

While it works for `dir` (quotes are removed), it fails for `user`:

```julia
Enter value for 'user' (required): Enter value for 'dir' (default: /tmp/jl_lgEt7r/dev): ┌ Warning: Quotes
│   t1.user = "\"me\""
│   t1.dir = "/home/gdalle"
│   t2.user = "me"
│   t2.dir = "/home/gdalle"
└ @ Main ~/Work/GitHub/Julia/PkgTemplates.jl/test/interactive.jl:200
```

Any idea why this happens?

PR #444, opened 2023-12-15
Mark broken interactive test as broken

gdalle:
Partial answer for #434 until we fix it

### DIFF
--- a/test/interactive.jl
+++ b/test/interactive.jl
@@ -199,7 +199,7 @@ end
             if get(ENV, "CI", "false") == "true"
                 @test_broken result
             else
-                @test result
+                @test_broken result
             end
 
             print(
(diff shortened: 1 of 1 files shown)


oxinabox:
The failures on 1.7.2 are real
https://github.com/JuliaCI/PkgTemplates.jl/actions/runs/7219646408/job/19671050381?pr=444#step:7:2870

test/interactive.jl
```diff
@@ -199,7 +199,7 @@ end
             if get(ENV, "CI", "false") == "true"
                 @test_broken result
             else
-                @test result
+                @test_broken result
```
oxinabox, line 202:
```suggestion
                @test_broken result  # see https://github.com/JuliaCI/PkgTemplates.jl/issues/434
```

oxinabox (approved):

PR #453, opened 2024-01-17
Run reference tests on 1.10 instead of 1.7

gdalle:
How I proceeded (might be useful for future updates)

1. change `test/runtests.jl` to run reference tests on v1.10.0
2. change `.github/workflows/CI.yml` to ensure that v1.10.0 is part of CI
3. go to `test/reference.jl`, comment out the following block and replace it with `copy_file(comparison, reference)`

```julia
    if PROMPT
        while true
            println("Update reference file? [y/n]")
            answer = lowercase(strip(readline()))
            if startswith(answer, "y") || startswith(answer, "n")
                # If the user chooses to update the reference file,
                # replace its contents with the comparison file.
                startswith(answer, "y") && copy_file(comparison, reference)
                break
            end
        end
    end
```

4. run the tests
5. revert the changes to `test/reference.jl`

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -21,6 +21,7 @@ jobs:
       matrix:
         version:
           - "1.0"  # LTS
+          - "1.7"  # interactive tests require <= 1.8
           - "1"    # Latest Release
         os:
           - ubuntu-latest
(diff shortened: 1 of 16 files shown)


gdalle:
The only things that changed were CI yml files, so maybe there's a more clever way than manual updating with each new stable Julia

oxinabox (approved):

gdalle:
I think we need to change the required checks before merging, which I don't have the rights to do (GitHub is still waiting on the tests to pass for 1.7.2). Can you do it or grant me permission?

oxinabox:
I have bumped your permissions all the way up. so you should now have them

PR #454, opened 2024-01-18
Switch default julia compat to 1.6.7

gdalle:
Most interesting changes:

- `src/template.jl`
- `src/plugins/ci.jl`
- `test/interactive`
- `test/show`

I still need advice on how to modify the interactive tests. For now we can only pick a 1.x.0, and it would be nice to have a default 1.6.7, right?

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -21,6 +21,7 @@ jobs:
       matrix:
         version:
           - "1.0"  # LTS
+          - "1.7" # interactive tests require <= 1.8
           - "1"    # Latest Release
         os:
           - ubuntu-latest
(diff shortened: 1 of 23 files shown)


gdalle:
This one is good to go. I haven't checked why the order of Julia versions changes in CI files, but if it's a problem I can fix it

oxinabox:
I'm fine with this. Though rumour has it that 1.10 is being considered for new LTS.
But we will need to keep testing something <=1.7 for interactive tests anyway.

oxinabox (approved):
I am fine with this, though rumour has it 1.10 is being considered for new LTS

gdalle:
> I am fine with this, though rumour has it 1.10 is being considered for new LTS

It probably won't be 1.10.0 though. And we have no timeline on this (probably close to 1.11 release). So this change still makes sense in the meantime

PR #460, opened 2024-02-09
Add Documenter key and warning about SSH setup

### LINKED ISSUE #459: Add Documenter key
And more generally update to the latest recommended documenter yml

https://documenter.juliadocs.org/stable/man/hosting/#GitHub-Actions

ping @goerz @adrhill if you wanna give it a shot

gdalle:
Fix #459


oxinabox (approved):
I wish we could automate this step

gdalle:
Me too, but I'm not sure we can fiddle with the GitHub API, especially with secret SSH keys, and live to tell the tale

adrhill:
> I wish we could automate this step

Maybe a solution between *just having a note in the docstring* and *calling the GitHub API* would be to print the output of `DocumenterTools.genkeys` from PkgTemplates?

PR #465, opened 2024-02-23
Allow docs/make.jl to be run from the root of the repo

MarkNahabedian:
Add the repo root to LOAD_PATH so that docs/make.jl can be run from the repo root as well as from the docs directory.

### DIFF
--- a/templates/docs/make.jl
+++ b/templates/docs/make.jl
@@ -1,3 +1,5 @@
+insert!(LOAD_PATH, 1, joinpath(@__DIR__, ".."))
+
 using {{{PKG}}}
 using Documenter
 


oxinabox:
I am not sure about this.
It is one way to do it.
But it's not the only way.

I am much more used to doing `] dev..` and checking in the docs/Manifest.

Or not checking in the manifest and doing that just prior to building.

gdalle:
Also this changes the global Julia state following docs build which is not ideal

MarkNahabedian:
My goal is that the current directory not affect the running of docs/make.jl.

I guess it might make more sense to do

```
cd(@__DIR__)
```

instead.

As for affecting global Julia state, my admittedly inelastic mind can't forsee a circumstance wehre make.jl is run from other than a stand-alone Julia session created solely for that purpose.  Please enlighten me.

Issue #466, opened 2024-03-26
compat entry for Test


putianyi889:
The auto-generated Project.toml doesn't pass Aqua test:
```
PkgName [uuid] does not declare a compat entry for the following extras:
1-element Vector{Base.PkgId}:
 Test [8dfed614-e22c-5e08-85e1-65c5234f0b40]
```

gdalle:
Good catch! Do you want to try and fix it in a PR? I can help

putianyi889:
> Good catch! Do you want to try and fix it in a PR? I can help

I'd like to, but I'm not sure if I'll need to change the definition of `Template` as it stores the Julia version.

oxinabox:
I think when you enable Aqua as a test argument in it also fails for Aqua itself

gdalle:
No, I think it just requires editing the `Test` plugin. Two cases to consider: separate `Project.toml` file or `[extras]` section in the main one. I think in both cases, you should be able to set the version for `Test` with `Pkg.compat`: https://pkgdocs.julialang.org/v1/api/#Pkg.compat

The relevant lines to edit are here:

https://github.com/JuliaCI/PkgTemplates.jl/blob/805bb7cfe158ab92ba76af3f2af5d7225f150714/src/plugins/tests.jl#L102-L153

PR #467, opened 2024-03-31
Add compat entry for Test

putianyi889:
- #466


oxinabox:
Does having a Test compat entry of 1 work on Julia 1.0 or on 1.6?
I recall something about compat values being funny from before we required them for StdLibs

gdalle:
I'm not sure there is a universally accepted answer, since they rolled back the registry requirements for stdlib compat: https://discourse.julialang.org/t/psa-compat-requirements-in-the-general-registry-are-changing/104958#update-november-9th-2023-2

So it is indeed a bit weird that Aqua still checks. Maybe we should align on what Aqua does?

Also related:

- #454

PR #472, opened 2024-05-17
github template: Bump julia-actions/setup-julia to v2

musoke:
I just noticed that the github action template still uses `julia-actions/setup-julia@v1`.  Bump that to v2.


gdalle:
Thanks! While we're at it, do you mind doing the same with the `cache` action?

gdalle (approved):

gdalle:
thanks for the contribution!

adrhill:
Can we tag a release for this?

PR #476, opened 2024-07-10
Switch default version to 1.10.4

oxinabox:
As of a few hours ago 1.10.4 is the LTS

not sure how much this breaks in the tests.
I bet it does, but i hope it doesn't
Will follow up to fix those if needed

Not sure if should also update our CI setup


gdalle:
Indeed 1.10 will be the LTS but it's not the case yet and it won't be as long as 1.11 hasn't been released, right?

oxinabox:
idk, last time LTS was changed it was indeed not til the release after it stopped being stable.
But the announcement at state of julia sounded like it was being made the LTS now.

1.11 is just around the corner anyway.

cvigilv:
Julia's LTS now its 1.10, any intention on revisiting this PR?

PR #477, opened 2024-07-19
feat: ignore all Manifest files

### LINKED ISSUE #462: add Manifest-v*.*.toml to default gitignore
I think this makes sense.
If we are ignoring Manifest.toml by default we should also ignore the `Manifest-v*.*.toml` that are introduced by  JuliaLang/julia#43845
OTOH one does not create these by accident -- they do have to be created via manual rename. So maybe we should leave it to the user?

frankebel:
Starting with Julia 1.11, version-specific `Manifest.toml` files can exist side-by-side.

Closes #462

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,6 @@
 *.jl.*.cov
 *.jl.cov
 *.jl.mem
-/Manifest.toml
-/test/fixtures/*/test/Manifest.toml
-/docs/Manifest.toml
+/Manifest*.toml
+/test/fixtures/*/test/Manifest*.toml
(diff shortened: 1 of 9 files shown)


oxinabox:
This needs fixing on 1.10.6
probably some reference files need to be regenerated

frankebel:
> This needs fixing on 1.10.6 probably some reference files need to be regenerated

Updated the reference files. Also changed the code to ignore `/docs/Manifest*.toml`.

oxinabox:
thank you. sorry for the the delay.

PR #479, opened 2024-08-18
Underscore was forgotten in variable name

Eben60:


gdalle (approved):
Thanks for spotting this!

PR #481, opened 2024-08-23
Use .jlt instead of .jl for template files

fredrikekre:
This patch changes the file extension of Julia template files from .jl to .jlt. The motivation for this is tooling that assume these are (valid) julia source files based on the .jl extension.


oxinabox:
I think this makes sense.
Most syntax highlighting tools etc (that are ok with broken code) still are not that useful  on the template files.
And the language can always be set manually.

Do you have merge rights?

oxinabox:
This CI error is real: 
https://github.com/JuliaCI/PkgTemplates.jl/actions/runs/10521345275/job/29151850911?pr=481#step:7:137

fredrikekre:
> This CI error is real:

Fixed, didn't show up locally (on the same Julia version).

> Do you have merge rights?

Looks like it.

oxinabox (approved):

oxinabox:
LGTM
Probably bump as a nonbreaking change so you can tag a release after you merge

PR #483, opened 2024-08-28
Func to fixup existing pkg

pat-alt:
Trying to reanimate #288 here to add support for a `fixup` function. I've tried to address the suggestions made [here](https://github.com/JuliaCI/PkgTemplates.jl/issues/16#issuecomment-636027919):

> I'd kind of like to not rely on Git. Obviously > 90% of packages generated with PkgTemplates are also Git repos but it's not a guarantee. I think maybe we could just back up the current state to a temp directory, print out a message saying "hey we're about to potentially overwrite a bunch of files, they're all backed up here, do you want to proceed?", and then if yes just write the files ignoring any current contents. Git users can then look at their diffs to see what changed. Maybe there could also be some check for "if it's already a Git repo, require that it's clean".

## Additions

1. The `fixup` function now backs up the current state of the package in a temporary directory, instead of solely relying on Git. 
2. A similar approach is taken to deal with an already existing README file: the old file is maintained and a new file that complies with the template used in `fixup` is added with a `_fixed.md` suffix. The user will receive an info message that they may need to move contents from the existing README into the new one and then overwrite the existing file with the new file. 
3. Other changes are just formatting (`JuliaFormatter.format("src")`).

Point (2) is clearly not optimal: ideally `fixup` would overwrite/add whatever is necessary to comply with the new template and keep existing content. I am not sure this can be done though and personally think it would be better to have mediocre `fixup` functionality than none at all?


gdalle:
Thank you for these contributions!
I just want to point out that @oxinabox and I chatted at JuliaCon, and agreed that [BestieTemplate.jl](https://github.com/abelsiqueira/BestieTemplate.jl) might be the best way forward to replace PkgTemplates.jl, precisely because it includes an update mechanism from the get go. We haven't made any formal announcement yet but I encourage you to check it out before you spend too much effort here.

PR #486, opened 2024-11-11
support JULIA_REFERENCETESTS_UPDATE env var in tests

simeonschaub:
to disable the interactive prompt as it was somehow broken for me locally. Also bumps the Julia version for reference tests to 1.10.6

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -43,7 +43,7 @@ jobs:
           # require this specific job to pass on all PRs
           # (see Settings > Branches > Branch protection rules).
           - os: ubuntu-latest
-            version: 1.10.0
+            version: 1.10.6
             arch: x64
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


oxinabox:
I wonder if we should use the enviroment variable
`JULIA_REFERENCETESTS_UPDATE`
Which is what is used by ReferenceTests.jl (Ideally we would just be using ReferenceTests.jl)

oxinabox (approved):

oxinabox (approved):

oxinabox:
I have invited you to be a collaborator on the package.
Once you accept feel free to merge when happy.
Usual ColPrac stuff applies: don't merge your own PR without someone (even if they don't have commit rights) approving it.

simeonschaub:
Cool, thanks!

simeonschaub:
Ah, I don't think I can change repo settings as a collaborator so someone else will have to change the required job to 1.10.6 first

PR #493, opened 2024-12-28
Edit TagBot docstring

Eben60:
Presumably an "or" was missing, thus changing the sense of the sentence into the opposite.


oxinabox:
thanks

PR #495, opened 2025-01-21
fix #494: interactive use of runic plugin

### LINKED ISSUE #494: Runic plugin not fully defined
Hi!

If the Runic plugin is selected interactively, the interactive creation of the package breaks as soon as it gets to the part of trying to configure Runic.

The stacktrace says that there is no method defined for `interactive(T::Type{Runic})`.

simeonschaub:


oxinabox (approved):
I assume you have checked this manually?

simeonschaub:
Yes, I confirmed it locally

PR #496, opened 2025-01-22
Add compat for Mocking 0.8

jaakkor2:
Test pass locally with Julia 1.10.7.  Some tests were skipped that would be run on earlier versions.


oxinabox:
Can you bump the patch version number and then i can merge this

oxinabox (dismissed):

PR #497, opened 2025-03-25
Add `repl_history.jl` to default ignore

timholy:
If you start Julia with

    $ JULIA_HISTORY=repl_history.jl julia --project

then the session history gets stored locally in the project directory
rather than your global `logs` folder. It's very unlikely that you want
to commit this file.

It might make some sense to ignore this file by default. On the other hand, since you can choose any name for it, perhaps this doesn't make much sense. I thought I'd submit the PR for consideration, but if the maintainers would rather just close it without merging, that's fine with me.

### DIFF
--- a/src/plugins/git.jl
+++ b/src/plugins/git.jl
@@ -2,7 +2,7 @@ const DEFAULT_DEFAULT_BRANCH = "main"
 
 """
     Git(;
-        ignore=String[],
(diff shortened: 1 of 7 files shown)


oxinabox:
I think because the file can have any name, it is better to put it in your individual global `.gitignore`
Because different users will put in different things.

PR #498, opened 2025-04-01
new release

oxinabox:


oxinabox:
I should have added this to #477 before merging it.
I didn't realize we had branch protection set high enough to deny me

PR #502, opened 2025-05-25
TagBot dispatch for custom registries

timholy:
When using plugin `TagBot(; registry="Foo/FooRegistry")`, support TagBot
dispatch on registry events.

CC @kdw503


oxinabox:
woah, very cool.
Is this documented anywhere?
I see the other end was added to the HolyLabRegistry in https://github.com/HolyLab/HolyLabRegistry/pull/161

timholy:
Let's rope in @IanButterworth

PR #515, opened 2025-12-17
add Downgrade CI

### LINKED ISSUE #509: Add Downgrade CI
[Something like this](https://github.com/SciML/SciMLBase.jl/blob/v2.24.0/.github/workflows/Downgrade.yml). I didn't see this in the docs, did I miss it? If not, is this something that would be welcomed before I begin on a PR?

kylebeggs:
I *think* I did this correctly with the help of Claude explaining to me, but I'm still not 100% on how this package works. Please suggest edits!

closes #509


src/plugins/ci.jl
```diff
@@ -417,7 +470,7 @@ function collect_versions(t::Template, versions::Vector)
     return sort(unique(vs))
 end
 
-const AllCI = Union{AppVeyor, GitHubActions, TravisCI, CirrusCI, GitLabCI, DroneCI}
+const AllCI = Union{AppVeyor, GitHubActions, TravisCI, CirrusCI, GitLabCI, DroneCI, DowngradeCI}
```
oxinabox: this isn't really a CI system like the others.
I do not think it be belongs here or in this file.

It's more its own thing. like the thing for building documentation.

I suggest also renaming it: to like "DowngradeDepencencyTests"

src/plugins/downgrade.jl
```diff
@@ -0,0 +1,51 @@
...
+
+function view(p::DowngradeDependencyTests, t::Template, pkg::AbstractString)
+    v = Dict(
+        "PKG" => pkg,
+        "USER" => t.user,
+        "SKIP" => join(p.skip, ","),
+        "HAS_SKIP" => !isempty(p.skip),
+    )
+    git = getplugin(t, Git)
+    if git !== nothing
```
oxinabox: should we error without this?

kylebeggs: I don't think so. Other plugins in this package don't error for similar situations.

kylebeggs: Good catch. I switched to the same pattern used by `TravisCI`, `GitLabCI`, `Codecov`, and `Coveralls`:

```julia
"BRANCH" => something(default_branch(t), DEFAULT_DEFAULT_BRANCH),
```

This always populates `BRANCH` (falling back to `"main"` when the `Git` plugin is absent) instead of silently omitting it. Erroring would have been inconsistent with every other CI/coverage plugin in the package, so this felt like the closer fit. Pushed in ed2b8f8.

kylebeggs:
im not convinced the last commit adding that test is any good...

PR #516, opened 2025-12-19
Replace CompatHelper with Dependabot as default for Julia dependencies

IanButterworth:
Dependabot now officially supports Julia package ecosystems: https://github.blog/changelog/2025-12-16-dependabot-version-updates-now-support-julia/

This PR replaces CompatHelper as the recommended approach for keeping package dependencies up to date.

Changes:
- Remove CompatHelper from default plugins
- Update Dependabot to monitor /, /docs, and /test directories
- Add Julia package ecosystem to dependabot.yml template
- Add commented-out grouping options for both ecosystems
- Update documentation to reflect the change
- Mark CompatHelper as deprecated (still available for users who prefer it)
- Update test fixtures accordingly

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -51,7 +51,7 @@ jobs:
         with:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
-      - uses: actions/cache@v4
+      - uses: actions/cache@v5
         env:
           cache-name: cache-artifacts
         with:
--- a/.github/workflows/JuliaNightly.yml
+++ b/.github/workflows/JuliaNightly.yml
@@ -13,7 +13,7 @@ jobs:
         with:
           version: nightly
           arch: x64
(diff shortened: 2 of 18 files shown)


oxinabox (approved):
It would be ideal if the Dependabot plugin took arguments for as to if to use if for GithubActions or Julia or both. (maybe 2 boolean flags).
I think it would be a useful enhancement to this -- often for personal projects I might not want to update my julia dependency ranges -- since old vesions keep working, but i do know i have to take updates on github-actions since a lot of the older versions of those break as endpoints get taken down.

With that said this can be fixed in post by editting the field by hand, so i don't think it is blocking -- we can just open an issue for this

DilumAluthge:
@oxinabox Is this good to merge (once the merge conflicts have been resolved)?

oxinabox:
>  Is this good to merge (once the merge conflicts have been resolved)?

Yes

PR #524, opened 2026-03-16
fix(license): clarify interactive license names

### LINKED ISSUE #448: License ambiguity
I always forget what the various licenses are from this list:

```
Enter value for 'destination' (default: LICENSE):
Select a license:
   MIT
   AGPL-3.0+
   ASL
   BSD2
   BSD3
   EUPL-1.2+
   GPL-2.0+
   GPL-3.0+
   ISC
   LGPL-2.1+
   LGPL-3.0+
 > MPL
```

Could perhaps more details be added to clarify what these are?

For example, if I google `ASL`, I see some people use it to mean "Academic Source License": https://github.com/ACEsuit/ACE.jl/blob/main/ASL.md. So I almost raised an issue asking about where the Apache one is, until I found https://github.com/JuliaCI/PkgTemplates.jl/issues/90 which explained things. Could those descriptions be added back?

Cheers,
Miles

Kemalau:
Closes #448

## Why
The interactive license prompt currently shows bare abbreviations such as `ASL` and `MPL`, which makes it easy to misread the intended license. This change keeps the saved license keys unchanged while showing expanded names in the interactive menu.

## What changed
- render the license picker entries as `abbreviation - full name`
- keep the existing return values and ordering intact
- extend the interactive test to assert the clearer prompt text

## Verification
- static review of the focused diff
- local execution of Julia tests was not possible in this environment because `julia` is not installed (`julia: command not found`)

### DIFF
--- a/src/plugins/license.jl
+++ b/src/plugins/license.jl
@@ -45,10 +45,30 @@ function prompt(::Type{License}, ::Type, ::Val{:name})
     # Move MIT to the top.
     deleteat!(options, findfirst(==("MIT"), options))
     pushfirst!(options, "MIT")
-    menu = RadioMenu(options; pagesize=length(options))
-    println("Select a license:")
+    menu = RadioMenu(map(license_display_name, options); pagesize=length(options))
+    println("Select a license (abbreviation - full name):")
     idx = request(menu)
     return options[idx]
 end
 
 customizable(::Type{License}) = (:name => String,)
+
+function license_display_name(name::AbstractString)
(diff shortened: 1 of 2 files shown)


oxinabox:
I think we should come back to this after doing
https://github.com/JuliaCI/PkgTemplates.jl/issues/448#issuecomment-4071841522

oxinabox:
If we think we still want this now that we are using SPIX names, can you rebase this PR?

PR #525, opened 2026-03-17
Use SPDX identifiers for license templates

### LINKED ISSUE #448: License ambiguity

MilesCranmerBot:
Fixes #448.

- Rename bundled license templates to SPDX license identifiers (e.g. `Apache-2.0`, `BSD-3-Clause`, `GPL-3.0-or-later`).
- Keep legacy names (e.g. `ASL`, `BSD3`, `GPL-3.0+`) working via a deprecated alias map.


oxinabox:
The tests probably need updating too

MilesCranmer:
@MilesCranmerBot can you check if the tests need to be updated and fix them if so?

(it only responds to tags from me - sorry for the noise)

oxinabox:
I seems there are no tests to be updated because the tests use ISC and MIT as examples and those haven't been renamed.

test/plugin.jl
```diff
@@ -107,4 +109,16 @@ PT.user_view(::FileTest, ::Template, ::AbstractString) = Dict("X" => 1, "Z" => 3
         # Only the final suffix is removed---we don't correct for user error
         @test pkg_name("Whee.jl.jl") == "Whee.jl"
     end
+
+    @testset "License aliases" begin
+        # Backward-compatibility: legacy license names should still work, but emit a depwarn.
+        apache = PT.default_file("licenses", "Apache-2.0")
+        @test isfile(apache)
```
oxinabox: I would write this as

```suggestion
        @assert isfile(apache)
```
as it isn't part of what we are testing.
But good work bot.

MilesCranmer: (@MilesCranmerBot can you integrate this suggestion please^)

PR #534, opened 2026-06-23
Ci actions autoupdate

Eben60:
# Automate GitHub Actions Versioning in Templates via Dependabot

## Main Goal
Currently, the GitHub Actions versions embedded in PkgTemplates's generated `CI.yml` templates are hardcoded. This causes them to quickly become outdated, forcing maintainers to manually bump them. 
This PR completely automates this process by delegating the version tracking to Dependabot. When Dependabot updates an action in PkgTemplates's own CI, those updated versions are dynamically injected into the generated package templates.

### 1. Dynamic Updates via Mustache Templating
All hardcoded action versions in `templates/github/workflows/CI.yml` (e.g., `actions/checkout@v2`) have been replaced with Mustache variables (e.g., `<<& actions/checkout>>`). 
- Introduced `src/parse_ci_actions.jl` to statically parse both `.github/workflows/CI.yml` and the newly introduced `.github/workflows/TriggerDependabotUpdate.yml`, extract the `uses:` versions, and expose them to the Mustache view.

### 2. Fixing `test/reference.jl` Issue
* **Infinite Loop Bug**: Fixed a bug discovered in the course of development of this PR when running the test suite via the REPL in a separate Terminal session. Apparently where stdin isn't fully interactive, `readline()` may repeatedly return an empty string (EOF), causing an infinite prompt loop. It now safely detects EOF and breaks out.

### 3. CI Architecture Changes
To automate the updates, including the updates of the reference tests fixtures, several interlocking workflows were introduced:

* **"Fake Workflow" as Action Versions Tracker**: Created `.github/workflows/TriggerDependabotUpdate.yml`. This workflow is designed to never run automatically. Its sole purpose is to list actions that are not present in `CI.yml` and *only* used in templates (like `julia-docdeploy` or `runic-action`) so Dependabot detects and updates them.
* **Auto-Updating Fixtures**: Created `.github/workflows/DependabotFixturesUpdate.yml`. This workflow runs *only* on Dependabot PRs. It runs the test suite with `JULIA_REFERENCETESTS_UPDATE="true"`. If the fixtures change, it automatically commits them back to the PR and uses the GitHub CLI to post a ⚠️ Warning comment, alerting maintainers to review the generated diffs.
* **Skipping Reference Tests for Dependabot**: When Dependabot bumps an action, the generated template changes, which breaks the reference tests in `test/fixtures/`. To prevent the main `CI.yml` from turning red, we introduced the `PT_SKIP_REFERENCE_TESTS` environment variable. If `github.actor == 'dependabot[bot]'`, the main CI cleanly skips the reference tests and passes. Should anything go wrong with the auto-update, it will be detected in a subsequent CI run e.g. on merging.
* **Broken Workflow Workaround**: While `test/reference.jl` already had a mechanism to auto-update fixtures via the `JULIA_REFERENCETESTS_UPDATE` environment variable, it has been intentionally designed to still fail the test (`@test :reference == :comparison`) after updating. This caused our new auto-updater workflow to crash abruptly. We patched this behavior so that when the environment variable is set on a CI runner, it issues a warning but otherwise passes the test, allowing the workflow to successfully commit the changes.
* **Manual Triggering Enabled**: In `CI.yml`, `on:   workflow_dispatch:` has been added. It was used during PR development, and I left it there, as it could be useful and otherwise does no harm.

## How to Test / Reproduce
If you want to test the entire Dependabot-triggered workflow on your own fork:
1. Ensure GitHub Actions are enabled in your fork (workflows are disabled by default in forks).
2. Ensure Dependabot is enabled in your fork's settings (`Settings -> Code security and analysis -> Dependabot version updates`).
3. Locally downgrade any action in `.github/workflows/CI.yml` or `TriggerDependabotUpdate.yml` (e.g., change `actions/checkout@v6` to `v4`). Commit and push to your default branch.
4. Force Dependabot to run immediately: Go to `Insights -> Dependency graph -> Dependabot -> Recent update jobs` and click **Check for updates**.
5. Dependabot will open a PR. You will see the main CI pass, the secondary auto-updater run, a new commit pushed with the updated `test/fixtures/`, and a warning comment placed on the PR!

**AI usage disclosure**: About 50% of the code changes have been generated by Gemini Pro 3.1 (High) within Antigravity. All human-produced code has been reviewed by Gemini and vice versa. All complex code changes have been additionally reviewed by Opus 4.6 in Antigravity.

P.S. As the reference fixtures are anyway updated in this PR, should we probably also bump the reference Julia version to the current LTS (i.e. 1.10.11)?

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,5 +1,5 @@
 name: CI
-# Run on master, tags, or any pull request
+# Run on master, tags, any pull request, and manual triggering
 on:
   schedule:
     - cron: '0 2 * * *'  # Daily at 2 AM UTC (8 PM CST)
@@ -7,6 +7,7 @@ on:
     branches: [master]
     tags: ["*"]
   pull_request:
+  workflow_dispatch:
 concurrency:
   # Skip intermediate builds: always.
   # Cancel intermediate builds: only if it is a pull request build.
@@ -16,6 +17,8 @@ jobs:
   test:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
     runs-on: ${{ matrix.os }}
+    env:
+      PT_SKIP_REFERENCE_TESTS: ${{ github.actor == 'dependabot[bot]' && 'true' || 'false' }}
     strategy:
       fail-fast: false
       matrix:
@@ -60,7 +63,7 @@ jobs:
             os: macOS-latest
             arch: aarch64
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@v7
       - uses: julia-actions/setup-julia@v3
         with:
           version: ${{ matrix.version }}
@@ -92,7 +95,7 @@ jobs:
     name: Documentation
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@v7
       - uses: julia-actions/setup-julia@v3
         with:
           version: '1'
--- a/.github/workflows/DependabotFixturesUpdate.yml
+++ b/.github/workflows/DependabotFixturesUpdate.yml
@@ -1,0 +1,67 @@
+name: Update Fixtures on Dependabot PR
+
+on:
+  pull_request:
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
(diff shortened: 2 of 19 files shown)


oxinabox:
Oh this is a very cool idea.
It is going to be a while before i can review it.

